PR #15, opened 2021-07-15
Going through all the documentation and trying to use

CGMossa:
I really, really appreciate this package. And I wanted to go through it and
use the new way to refer to types in rustdoc. A few things have come up, that
would be nice if we could go through:

### DIFF
--- a/examples/default_array.rs
+++ b/examples/default_array.rs
@@ -13,7 +13,7 @@ fn main() {
 fn matrix<const N: usize>() -> [[usize; N]; N] {
     let mut matrix = [[0; N]; N];
 
-    #[allow(clippy::clippy::needless_range_loop)]
+    #[allow(clippy::needless_range_loop)]
     for i in 0..N {
         for j in 0..N {
             matrix[i][j] = (i + 1) * (j + 1);
--- a/papergrid/src/lib.rs
+++ b/papergrid/src/lib.rs
@@ -181,7 +181,7 @@ impl Grid {
         self.size.0 += 1;
     }
 
-    /// Remove_row removes a `row` from a grid.
+    /// Removes a `row` from a grid.
     ///
     /// The row index must be started from 0
     pub fn remove_row(&mut self, row: usize) {
@@ -190,7 +190,7 @@ impl Grid {
         self.size.0 -= 1;
     }
 
-    /// Remove_row removes a `column` from a grid.
+    /// Removes a `column` from a grid.
     ///
     /// The column index must be started from 0
     pub fn remove_column(&mut self, column: usize) {
--- a/src/alignment.rs
+++ b/src/alignment.rs
@@ -1,4 +1,6 @@
-use crate::{CellOption, Table};
+use crate::CellOption;
+#[allow(unused)]
+use crate::Table;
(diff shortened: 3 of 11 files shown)


CGMossa:
Also, I don't really understand the `ChangeRing`: https://github.com/CGMossa/tabled/blob/d40f8307ab00f85ea5d02b4a2ff6708421d9c99f/src/lib.rs#L204-L205

zhiburt:
> Also, I don't really understand the ChangeRing: https://github.com/CGMossa/tabled/blob/d40f8307ab00f85ea5d02b4a2ff6708421d9c99f/src/lib.rs#L204-L205

`ChangeRing` is an old setting which was removed in regard of `Format` setting.
https://github.com/zhiburt/tabled/commit/6573015fb8777550d85800f3ba1c7f2645c07e9a#diff-a373be809d5fed898f6aa3a559af2601b9bd1c81ecb2a77be2821c77976a1312L47

So yes it certainly a bit outdated doc comment. Good catch :+1:

zhiburt:
Overall I agree that documentation itself  deserve to be polished.

zhiburt:
> For instance this line here: https://github.com/CGMossa/tabled/blob/25c331008023fecce39736ef4e3ec2d0adbe6631/src/panel.rs#L5-L6

You can basically remove any row/column by `Disable` option (Look and ./README.md).
And I was thinking that Remove something after adding a row may cause issues but actually now I am thinking about it and it must be fine. I mean the comment should be deleted..

CGMossa:
Great! Thanks for the answer. I've removed these lines. Feel free to add any other suggestions, and merge it when you're happy with it.

zhiburt (changes requested):
I actually like all your changes :100: 

And I wasn't be aware that when something is in scope you can refer on it in doc just like `[something]` so thanks for the lesson :)

But this feature brings compiler warnings. Wouldn't it be better to follow old fashioned (not actually) approach like this `[Table]: ./struct.Table.html`?

```bash
warning: unused import: `Table`
 --> src/alignment.rs:1:25
  |
1 | use crate::{CellOption, Table};
  |                         ^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused import: `Table`
 --> src/disable.rs:1:30
  |
1 | use crate::{bounds_to_usize, Table, TableOption};
  |                              ^^^^^

warning: unused import: `papergrid::Grid`
 --> src/object.rs:6:5
  |
6 | use papergrid::Grid;
  |     ^^^^^^^^^^^^^^^

warning: unused import: `Disable`
 --> src/panel.rs:1:13
  |
1 | use crate::{Disable, TableOption};
  |             ^^^^^^^

warning: 4 warnings emitted
```

CGMossa:
Based on what I've read here: https://github.com/rust-lang/rust/issues/79542, I think we should just do with the `#[allow(unused)]`. 
I'm pushing a commit that does this now.

CGMossa:
The same issue showcases a "better" pattern that is also discouraged in the same thread:

```rust
/// [`HashMap`]
///
/// [`HashMap`]: std::collections::HashMap
pub fn foo() {}
```

zhiburt:
Thanks for your PR.
And more importantly for your attention to the DOC. I'll try to work on it in a few days.

PR #21, opened 2021-09-09
Expaned display

### LINKED ISSUE #20: Support psql-like 'expanded display'
psql has an 'expanded display' mode, which converts the format to two columns, with the header in the first column and value in the second, and one row printed at a time:

```
peter@localhost testdb=> \a \t \x
Output format is aligned.
Tuples only is off.
Expanded display is on.
peter@localhost testdb=> SELECT * FROM my_table;
-[ RECORD 1 ]-
first  | 1
second | one
-[ RECORD 2 ]-
first  | 2
second | two
-[ RECORD 3 ]-
first  | 3
second | three
-[ RECORD 4 ]-
first  | 4
second | four
```

(example taken from the bottom of https://www.postgresql.org/docs/current/app-psql.html).

It would be great if something similar could be supported by this crate, as it helps a lot when rows have so many columns that they span multiple lines.

Thanks for the crate by the way, it's working wonderfully!

zhiburt:
closes #20

### DIFF
--- a/src/display/expanded_display.rs
+++ b/src/display/expanded_display.rs
@@ -7,8 +7,10 @@ use crate::Tabled;
 ///
 /// It escapes strings to resolve a multi-line ones.
 /// Because of that `colors` may not be rendered.
+#[derive(Debug)]
 pub struct ExpandedDisplay {
     format_record_splitter: fn(usize) -> String,
+    format_value: fn(String) -> String,
     fields: Vec<String>,
     records: Vec<Vec<String>>,
 }
@@ -20,9 +22,10 @@ impl ExpandedDisplay {
         let header = T::headers();
 
         Self {
-            format_record_splitter: |i| format!("-[ RECORD {} ]-", i),
             records: data,
             fields: header,
+            format_record_splitter: |i| format!("-[ RECORD {} ]-", i),
+            format_value: |s| s,
         }
     }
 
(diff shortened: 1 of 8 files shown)


src/display/expanded_display.rs
```diff
@@ -0,0 +1,169 @@
...
+                .collect::<String>()
+                .escape_debug()
+                .to_string()
+        }));
+        self
+    }
+
+    /// Sets max width of value,
+    /// when limit is reached next chars will be placed on the next line.
+    pub fn format_value_max_width_wrapped(&mut self, max: usize) -> &mut Self {
```
sd2k, line 68: Could just be called `max_width_wrapped`? Although I agree, I don't really love the naming duplication here, it feels like it would be nice to have methods like:

```rust
impl ExpandedDisplay {
    pub fn max_width(&self, max: usize) -> &mut Self;
    pub fn truncate(&self, truncate: bool) -> &mut Self;
    pub fn wrap(&self, wrap: bool) -> &mut Self;
}
```
where calling `x.wrap(true)` would switch off truncation, and vice versa. But I think that would require generating the `self.format_value` attribute at `fmt`-time, rather than setting it as a boxed closure 🤔

src/display/expanded_display.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    /// Turn off a wrapping of multiline value.
+    pub fn format_value_in_one_line(&mut self) -> &mut Self {
+        self.format_value = Some(Box::new(|s| s.escape_debug().to_string()));
+        self
+    }
+
+    /// Sets max width of value.
+    /// The rest will be trunceted.
+    pub fn format_value_max_width(&mut self, max: usize) -> &mut Self {
```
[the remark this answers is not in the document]
sd2k: Also, do you think it could be worth adding a suffix if the value is truncated, like in the main `MaxWidth` case? (It would mean another argument to control that, though)

sd2k (approved):
This looks perfect for my use case at least, I've tested out all the options and they work great. I've tried to come up with a few suggestions for method names, in case they're useful 👍

src/display/expanded_display.rs
zhiburt: what about just having this?

```
impl ExpandedDisplay {
    pub fn truncate(&self, max: usize) -> &mut Self;
    pub fn wrap(&self, max: usize) -> &mut Self;
}
```

zhiburt: I renamed it as I propoused

sd2k: Nice, looks good!

PR #23, opened 2021-09-09
Modify 'MaxWidth' to allow wrapping as well as truncating contents

sd2k:
Unfortunately this changes the API of `MaxWidth` so that it's constructed using associated functions instead of all the contents being public - perhaps there's a way to do it without that though!

I've pulled in the unicode segmentation crate here to ensure we're not
splitting a string in the middle of a unicode grapheme (see
https://stackoverflow.com/questions/68122526/splitting-a-utf-8-string-into-chunks),
but I'm not sure that's being considered elsewhere in the crate atm -
perhaps this PR should use chars rather than graphemes, and grapheme
consideration can be added in a later PR.

Also this doesn't consider colours - need to think about how that should
be done for the string chunks.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ color = ["papergrid/color", "ansi-cut"]
 tabled_derive = {path = "./tabled_derive"}
 papergrid = "0.1.22"
 ansi-cut = { version = "0.1.0", optional = true }
+unicode-segmentation = "1.8.0"
 
 [dev-dependencies]
 owo-colors = "1"
--- a/examples/width.rs
+++ b/examples/width.rs
@@ -12,7 +12,7 @@ fn main() {
 
     let table = Table::new(&data)
         .with(Style::github_markdown())
-        .with(Modify::new(Full).with(MaxWidth(10, "...")));
+        .with(Modify::new(Full).with(MaxWidth::truncating(10, "...")));
 
     println!("{}", table);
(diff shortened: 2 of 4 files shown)


zhiburt:
> Unfortunately this changes the API of MaxWidth so that it's constructed using associated functions instead of all the contents being public - perhaps there's a way to do it without that though!

I guess it's fine.

> I've pulled in the unicode segmentation crate here to ensure we're not
splitting a string in the middle of a unicode grapheme (see
https://stackoverflow.com/questions/68122526/splitting-a-utf-8-string-into-chunks),
but I'm not sure that's being considered elsewhere in the crate atm -
perhaps this PR should use chars rather than graphemes, and grapheme
consideration can be added in a later PR.

Yes it's a complex topic I think its certainly shouldn't be addressed here.

> Also this doesn't consider colours - need to think about how that should
be done for the string chunks.

It must be possible to laverage [ansi_cut](https://docs.rs/ansi-cut/0.1.0/ansi_cut/trait.AnsiCut.html#tymethod.cut) we use it for truncate option already.
Even though it may be inefficient

zhiburt:
Here's a small refactoring https://github.com/zhiburt/tabled/commit/a8aab8fc9b5495c903a82432cbb717572a7909b2

It turns out isn't as straitforward to make a colored wrapping work.
I'll tackle it later.

PR #38, opened 2021-11-23
Switch from String backend to Cow<str>

zhiburt:
Change of `Tabled` trait to

```rust
trait Tabled {
    fn fields(&self) -> Vec<Cow<'_, str>>;
    fn headers() -> Vec<Cow<'static, str>>;
}
```

The idea that we don't waste memory by making copies of `&str`.

Though the interface became less versatile...

I guess there must be done some refactoring then.

### DIFF
--- a/papergrid/src/lib.rs
+++ b/papergrid/src/lib.rs
@@ -23,6 +23,7 @@
 //! ```
 
 use std::{
+    borrow::Cow,
     cmp::max,
     collections::HashMap,
     fmt::{self, Display},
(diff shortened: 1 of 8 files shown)


zhiburt:
Removing deprecated functions will be a breaking change.
So probably then the library must bump version to 1.0.0

zhiburt:
The idea of using Cow in paper grid is valuable but changes in Tabled trait are less attractive and require some pondering still.

Issue #31, opened 2021-11-18
Add a `join` method


zhiburt:
```rust
let table3 = table1.join(table2);
```

### Notes:

* ? Add an optional argument to define which stylies must be used
*  ? Add options to determine how to act if one table have a different amount of rows then the other

senk8:
Hi @zhiburt , I'm interested in this.
Can I work on this?

zhiburt:
Hi @senk8 

You're welcome.
Feel free to contact me if there was any questions.

I haven't noticed your comment in a first place,
I hope merge of #35 doesn't affected you :disappointed_relieved:

zhiburt:
I can imagine the implementation could be something similar to https://github.com/zhiburt/tabled/blob/d5aa14dd2e0a2cdfe149f6ce9be874de0db1acff/src/disable.rs#L58-L75

Here we delete row.
But in `join` case we need to add rows/columns.

And just to note that `papegrid::Grid` is intentionally immutable from perspective of rows/columns. We can't add any rows and columns dynamically. 
We can only change data and styles.

*There are a bunch of things that must be documented :smile:

zhiburt:
Note2: 

The method could be exposed via an object which implements `TableOption`.
This way we could set a different options.

Something like this.

```rust
table1.with(Join::inner(table2))
```

Not sure which is the best.
Just thoughts.

zhiburt:
Note3:

I guess we shouldn't be worried about styles.
As we do in `Dissable` and `Panel`.

The style could be changed after `join`.

So essentially the main thing seems to copy the data.

zhiburt:
Hey @senk8

Might you need some help?

senk8:
Thank you, @zhiburt!

```rust
let table3 = table1.join(table2);
```

I tried to join the two grids together, using `Grid::get_content` to copy only the data, but this way the style information of `table1` and `table2` is lost.  How should I determine the style by default?

senk8:
FYI 
https://github.com/senk8/tabled/blob/add-join-method/src/lib.rs#L266-L289

zhiburt:
> FYI
> https://github.com/senk8/tabled/blob/add-join-method/src/lib.rs#L266-L289

Looks good.

> I tried to join the two grids together, using Grid::get_content to copy only the data, but this way the style information of table1 and table2 is lost. In this case, how should the style be determined ?

Actually never was thinking that way.
I was thinking that we will use all styles.

If you change your inner loop to the following.
[...]

PR #40, opened 2021-12-08
Add join method

senk8:
It relates #31 . 

Note
- What options do I need?
- Is there a way to write the `pub mod` that doesn't require it to be specified?

### DIFF
--- a/src/join.rs
+++ b/src/join.rs
@@ -4,34 +4,14 @@ use crate::TableOption;
 use papergrid::{Entity, Grid};
 
 pub enum Join {
-    VerticalTop(Table),
-    VerticalBottom(Table),
-    HorizontalLeft(Table),
-    HorizontalRight(Table),
+    Vertical(Table),
+    Horizontal(Table),
 }
 
 impl TableOption for Join {
     fn change(&mut self, other: &mut Grid) {
         match self {
-            Join::VerticalTop(table) => {
-                let new_row_size = table.grid.count_rows() + other.count_rows();
-                let mut new_grid = Grid::new(new_row_size,table.grid.count_rows());
-
-                for column in 0..new_grid.count_columns() {
-                    for row in 0..new_grid.count_rows() {
-                        let settings = if row < table.grid.count_rows() {
-                            table.grid.get_settings(row, column)
-                        } else {
-                            let row = row - table.grid.count_rows();
-                            other.get_settings(row, column)
-                        };
-                        new_grid.set(&Entity::Cell(row,column),settings.border_restriction(false));
-                    }
-                }
-            
-                *other = new_grid;
-            },
-            Join::VerticalBottom(table) => {
+            Join::Vertical(table) => {
                 let new_row_size = table.grid.count_rows() + other.count_rows();
                 let mut new_grid = Grid::new(new_row_size,table.grid.count_rows());
 
@@ -49,25 +29,7 @@ impl TableOption for Join {
 
                 *other = new_grid;
             },
-            Join::HorizontalLeft(table) => {
-                let new_column_size = table.grid.count_columns() + other.count_columns();
-                let mut new_grid = Grid::new(table.grid.count_rows(),new_column_size);
-
-                for column in 0..new_grid.count_columns() {
-                    for row in 0..new_grid.count_rows() {
-                        let settings = if column < table.grid.count_columns() {
-                            table.grid.get_settings(row, column)
-                        } else {
-                            let column = column - table.grid.count_columns();
-                            other.get_settings(row, column)
-                        };
-                        new_grid.set(&Entity::Cell(row,column),settings.border_restriction(false));
-                    }
-                }
-            
-                *other = new_grid;
-            },
-            Join::HorizontalRight(table) => {
+            Join::Horizontal(table) => {
                 let new_column_size = table.grid.count_columns() + other.count_columns();
                 let mut new_grid = Grid::new(table.grid.count_rows(),new_column_size);
 
--- a/src/join.rs
+++ b/src/join.rs
@@ -14,18 +14,24 @@ enum JoinMode {
 }
 
 impl Join {
-  pub fn vertical(table: Table) -> Self {
-    Self { mode: JoinMode::Vertical(table), strict_size: false }
-  }
+    pub fn vertical(table: Table) -> Self {
+        Self {
+            mode: JoinMode::Vertical(table),
+            strict_size: false,
+        }
+    }
 
-  pub fn horizontal(table: Table) -> Self {
-    Self { mode: JoinMode::Horizontal(table), strict_size: false }
-  }
+    pub fn horizontal(table: Table) -> Self {
+        Self {
+            mode: JoinMode::Horizontal(table),
+            strict_size: false,
+        }
+    }
 
-  pub fn strict_size(mut self) -> Self {
-    self.strict_size = true; 
-    self
-  }
+    pub fn strict_size(mut self) -> Self {
+        self.strict_size = true;
+        self
+    }
 }
 
 impl TableOption for Join {
@@ -94,7 +100,7 @@ impl TableOption for Join {
                     for row in 0..table.grid.count_rows() {
                         let settings = table.grid.get_settings(row, column);
                         new_grid.set(
-                            &Entity::Cell(other.count_rows()+row, column),
+                            &Entity::Cell(other.count_rows() + row, column),
                             settings.border_restriction(false),
                         );
                     }
--- a/src/join.rs
+++ b/src/join.rs
@@ -1,0 +1,52 @@
+#[allow(unused)]
+use crate::Table;
+use crate::TableOption;
+use papergrid::{Entity, Grid};
+
+pub enum Join {
+    Vertical(Table),
+    Horizontal(Table),
+}
+
+impl TableOption for Join {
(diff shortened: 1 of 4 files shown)


zhiburt:
Looks good :100: 

I have a couple of notes though.

1. I think we need to handle a case where `data1.len() != data2.len()`.
Currently as I see we get panicked at 'index out of bounds: the len is 5 but the index is 5`.
    
    By default we could keep the such cells empty.
    But it should be optional. 
    
    Here's a tests that can be used
    
    ```rust
    #[test]
    fn table_join_horizontal_with_different_length() {
        #[derive(Tabled)]
        enum Domain {
            Security,
            Embeded,
            Frontend,
            Unknown,
        }
    
        #[derive(Tabled)]
        struct Developer(#[header("name")] &'static str);
    
        let data1 = vec![
            (Developer("Terri Kshlerin"), Domain::Embeded),
            (Developer("Catalina Dicki"), Domain::Security),
            (Developer("Jennie Schmeler"), Domain::Frontend),
            (Developer("Maxim Zhiburt"), Domain::Unknown),
        ];
    
        let table1 = Table::new(data1).with(Style::psql());
    
        let data2 = vec![
            (Developer("Taro Yamada"), Domain::Embeded),
            (Developer("Hanako Sato"), Domain::Security),
            (Developer("Ken Nakamura"), Domain::Unknown),
            (Developer("Hiro Takahashi"), Domain::Frontend),
            (Developer("Hiro Takahashi"), Domain::Frontend),
        ];
    
        let table2 = Table::new(data2).with(Style::psql());
        let table3 = table1.with(join::Join::HorizontalRight(table2));
    
        assert_eq!(
            table3.to_string(),
            concat!(
                "      name       | Security | Embeded | Frontend | Unknown       name      | Security | Embeded | Frontend | Unknown \n",
                "-----------------+----------+---------+----------+-------------------------+----------+---------+----------+---------\n",
                " Terri Kshlerin  |          |    +    |          |           Taro Yamada   |          |    +    |          |         \n",
                " Catalina Dicki  |    +     |         |          |           Hanako Sato   |    +     |         |          |         \n",
                " Jennie Schmeler |          |         |    +     |           Ken Nakamura  |          |         |          |    +    \n",
                "  Maxim Zhiburt  |          |         |          |    +     Hiro Takahashi |          |         |    +     |         \n",
                "                 |          |         |          |          Hiro Takahashi |          |         |    +     |         \n",
            )
        );
    }
    ```
    
    ```rust
    #[test]
    fn table_join_vertical_with_different_length() {
        #[derive(Tabled)]
        enum Domain {
            Security,
            Embeded,
            Frontend,
            Unknown,
        }
    
        #[derive(Tabled)]
        struct Developer(#[header("name")] &'static str);
    
        let data1 = vec![
            (Developer("Terri Kshlerin"), Domain::Embeded),
            (Developer("Catalina Dicki"), Domain::Security),
            (Developer("Jennie Schmeler"), Domain::Frontend),
            (Developer("Maxim Zhiburt"), Domain::Unknown),
        ];
    
        let table1 = Table::new(data1).with(Style::psql());
    
        let data2 = vec![
            (Developer("Taro Yamada"), Domain::Embeded),
            (Developer("Hanako Sato"), Domain::Security),
            (Developer("Ken Nakamura"), Domain::Unknown),
            (Developer("Hiro Takahashi"), Domain::Frontend),
            (Developer("Hiro Takahashi"), Domain::Frontend),
        ];
    
        let table2 = Table::new(data2).with(Style::psql());
        let table3 = table1.with(join::Join::VerticalTop(table2));
    
        assert_eq!(
            table3.to_string(),
            concat!(
                "      name       | Security | Embeded | Frontend | Unknown \n",
                "-----------------+----------+---------+----------+---------\n",
                "   Taro Yamada   |          |    +    |          |         \n",
                "   Hanako Sato   |    +     |         |          |         \n",
                "  Ken Nakamura   |          |         |          |    +    \n",
                " Hiro Takahashi  |          |         |    +     |         \n",
                " Hiro Takahashi  |          |         |    +     |         \n",
                "      name       | Security | Embeded | Frontend | Unknown \n",
                "-----------------+----------+---------+----------+---------\n",
                " Terri Kshlerin  |          |    +    |          |         \n",
                " Catalina Dicki  |    +     |         |          |         \n",
                " Jennie Schmeler |          |         |    +     |         \n",
                "  Maxim Zhiburt  |          |         |          |    +    \n",
            )
        );
    }
    ```

2. What is the point to have a Top/Bottom combination?
    An order of arguments could be changed to reach the same affect.
    I mean call ` table1.with(join::Join::Vertical(table2));` for Bottom style of vertical join or cal `table2.with(join::Join::VerticalTop(table1));` for top one.
   *Left/right is a bit different. Look at #31 I describe a new idea there.

zhiburt:
Overall I think you shouldn't be worry with this comment https://github.com/zhiburt/tabled/issues/31#issuecomment-989773210

We can merge your PR with the functionality you've already implemented and then iterate on it.

senk8:
@zhiburt 

> I think we need to handle a case where data1.len() != data2.len().
> Currently as I see we get panicked at 'index out of bounds: the len is 5 but the index is 5`.
> 
> In case of horizontal join we could keep the such cells empty.
> But it should be a configurable option.

How should it be designed?  Here are two ideas I have.

1. We add variant of `Join::StrictHorizontal(Table)`. The behavior is determined by whether variant is `StrictHorizontal` or `Horizontal`.

2. We modify the enum  to be  `Join::Horizontal(Table,bool)`. The behavior is determined by whether bool is `true` or `false`.

Please feel free to give me any ideas.

zhiburt:
Aaaa having it's as a optional behavior might be a good idea.

I would rather go with something like this.

```rust
pub struct Join {
  mode: JoinMode,
  strict_size: bool,
}

pub enum JoinMode {
    Vertical,
    Horizontal,
}
```

senk8:
https://github.com/zhiburt/tabled/pull/40#issuecomment-991946855
This is very cool 👍 . I am working on the design.

senk8:
@zhiburt 
Sorry for the late reply 😔.  I have implemented all the features except https://github.com/zhiburt/tabled/issues/31#issuecomment-989773210 . We didn't discuss the case of different column sizes in vertical join, what do you think?  I think that we can use empty cell like a horizontal join.

zhiburt:
> Sorry for the late reply pensive

That's fine I was actually wondering if you need some help :)

> We didn't discuss the case of different column sizes in vertical join, what do you think? I think that we can use empty cell like a horizontal join.

I think this is good.
Do you plan to implement it?

src/join.rs
```diff
@@ -0,0 +1,113 @@
...
+
+pub struct Join {
+    pub mode: JoinMode,
+    pub strict_size: bool,
+}
+
+pub enum JoinMode {
+    Vertical(Table),
+    Horizontal(Table),
+}
```
zhiburt, line 15: What do you think about this design?

```rust
pub struct Join {
    mode: JoinMode,
    strict_size: bool,
}

enum JoinMode {
    Vertical(Table),
    Horizontal(Table),
}

impl Join {
  pub fn vertical() -> Self {
    Self { mode: JoinMode::Vertical, strict_size: false }
  }

  pub fn horizontal() -> Self {
    Self { mode: JoinMode::Horizontal, strict_size: false }
  }

  pub fn strict_size(mut self) -> Self {
    self.strict_size = true; 
    self
  }
}
```

src/join.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl TableOption for Join {
+    fn change(&mut self, other: &mut Grid) {
+        match &mut self.mode {
+            JoinMode::Vertical(table) if self.strict_size => {
+                let new_row_size = table.grid.count_rows() + other.count_rows();
+                let mut new_grid = Grid::new(new_row_size, table.grid.count_rows());
```
zhiburt, line 22: It seems to me this is the only line which is different from the other `Vertical` branch?
It may be worth to combine these 2 to eliminate duplication?
I might be wrong though as I haven't thoroughly checked it.

https://github.com/zhiburt/tabled/blob/e66253b35bc17de83a80ec8f39e82c7c8c82064d/src/join.rs#L64

zhiburt (commented):
Overall seems good to me.
We can merge it already if you wish.

src/join.rs
senk8: LGTM 👍

src/join.rs
senk8: Since other.count_rows()is equal to cmp::max(other.count_rows(),table.grid.count_column()) here, I used other.count_rows() instead of new_row_size.
This may cause a misunderstanding like this time, so I'll fix it.

senk8:
I checked the reviews and fix according to review. If there are no problems, I would like to merge.

zhiburt:
I'll merge it today.
Thank you for working on it.
I hope it did not consume a lot of your energy and time.

Merry Christmas and happy New Year

closes #31

senk8:
I'm sorry, I don't think the formatter was applied. I've formatted it.

senk8:
Excuse me, Why is clippy failing in CI?  It doesn't seem to have anything to do with  changes of this PR.

zhiburt:
> Excuse me, Why is clippy failing in CI? It doesn't seem to have anything to do with changes of this PR.

It's not related to your PR

It's issues in master `branch` which appeared after update of clippy on CI.

PR #41, opened 2021-12-24
Update README.md [docs.rs] badge to show the latest version

panarch:


zhiburt:
Hi @panarch good catch.

Thank you for bringing the fix.

PR #42, opened 2021-12-26
Update readme about Style usage

24seconds:
Hello, I'm using tabled in [my pomodoro application](https://github.com/24seconds/rust-cli-pomodoro) and it's really cool!

While applying the Style, I found README about `Style` explanation is not matched to current version (as of [v0.3.0](https://docs.rs/tabled/0.3.0/tabled/style/struct.Style.html#impl)).
So I made update :D.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ A list of ready to use styles.
 Styles can be chosen by passing a `Style` argument option.
 
 ```rust
-let table = Table::new(&data).with(Style::PSQL);
+let table = Table::new(&data).with(Style::psql());
 ```
 
 ### Default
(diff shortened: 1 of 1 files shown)


zhiburt:
Hey @24seconds 

> While applying the Style, I found README about Style explanation is not matched to current version (as of v0.3.0).

That's true

But the `README.md` corresponds to `master` branch in which we now have `constants` like `Style::ASCII`.
*These changes not released yet as `0.4`

So I am not sure if your changes applicable.

But I am interesting in your opinion according to the design of `Style`.
Because I was thinking to actually remove methods like `Style::psql()` `Style::pseudo()` at all
(Currently they're marked as deprecated in `master`).
In regard of constants like `Style::PSQL`, `Style::Pseudo`.

From performance point of your its all the same.
We could keep both as well but it brings a bit of duplication.

So I am curious what you think about this.

24seconds:
Oh, got it. Thank you for the detailed explanation. For me, enum seems more comfortable. Let me close this pr!

Issue #36, opened 2021-11-22
Add a trait for `table()` on `Iterator`


zhiburt:
It seems possible to create a `trait` which would be implemented for any `Iterator`.
So the following calls could be done.

```rust
let people: &[Person] = ...
let table = people.table()
                  .with(Style::psql());
```

```rust
let people: Vec<Person> = ...
let table = people.table()
                  .with(Style::psql());
```

Is it seem like a good idea?

The idea of implementation can be taken from https://youtu.be/bnnacleqg6k

zhiburt:
closed in #43

PR #43, opened 2021-12-26
Implement table() trait

24seconds:
I watched the video in Issue #36 and it seems interesting. So I made `TableIteratorExt` trait to support `table()` function. 
(If the comment are needed let me add it. If the trait name is not good let me change it)

```rust
let people: Vec<Person> = ...
let table = people.table()
                  .with(Style::psql());

```
The `TableIteratorExt` applies to the type `U` which is bound to `IntoIterator<Item = T>` type. (T is bound to `Tabled` type)

I don't know which tests to be added, so I simply added one test to `table_test.rs`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -264,6 +264,20 @@ impl Table {
     }
 }
 
+pub trait TableIteratorExt {
+    fn table(self) -> Table;
+}
+
+impl<T, U> TableIteratorExt for U
+where
+    T: Tabled,
+    U: IntoIterator<Item = T>,
+{
+    fn table(self) -> Table {
+        Table::new(self)
+    }
+}
+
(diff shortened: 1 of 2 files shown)


zhiburt:
Hi @24seconds 

That looks amazing.

Might you add a doc comment to `TableIteratorExt`.
And some notion about it in `README.md`?

I can do it myself if you wish.

24seconds:
@zhiburt Thanks for the reply. I added comment to `TableIteratorExt`. Could you write notion about in `README.md`? I don't know which part would be proper. Of course, you can modify the comment I added :D.

zhiburt:
Thank you for your contribution.
I'll merge it today.

Marry Christmas and Happy New Year

PR #70, opened 2022-02-23
tabled/ Fix an issue with a panel rendering

### LINKED ISSUE #66: Fix `Panel` style rendering on tables with more than 2 columns
Running this code will result in the following.

```rust
use tabled::{Style, Table};

fn main() {
    let table = Table::new(&[(0, 1)])
        .with(tabled::Panel("Numbers", 0))
        .with(Style::modern());

    println!("{}", table);
}
```

```rust
┌───────────┬
│Numbers    │
├───────────┼
│ i32 │ i32 │
├─────┼─────┤
│  0  │  1  │
└─────┴─────┘
```

As you may see the styles on the right side is not exactly correct.
It would be good to be fixed.

zhiburt:
close #66

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ color = ["papergrid/color", "ansi-str"]
 
 [dependencies]
 tabled_derive = "0.2.0"
-papergrid = "0.2.1"
+papergrid = { path = "./papergrid" }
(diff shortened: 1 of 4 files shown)


zhiburt:
It does solves original issue.
But we would need to rething the styling a panels.

Issue #33, opened 2021-11-18
Add `extract` method


zhiburt:
```rust
// grid

+---+---+---+
|0-0|0-1|0-2|
+---+---+---+
|1-0|1-1|1-2|
+---+---+---+
|2-0|2-1|2-2|
+---+---+---+

// grid.extract(.., ..1)

+---+
|0-0|
+---+
|1-0|
+---+
|2-0|
+---+
```

Can this be useful?

The method itself is already in `papergrid`.

https://github.com/zhiburt/tabled/blob/2cdda738b8612d4fb71d29c84123ed590a494315/papergrid/tests/extract.rs

zhiburt:
closed via #87

PR #87, opened 2022-03-15
Extract

IsaacCloos:
#33 
Hello, this is my first public pull request. Any feedback would be greatly appreciated 🙂

This change would add a new Table api feature `Extract`. This feature would allow user to create a new Table from a segment of the original to focus on a reduced number of rows and columns.

Todo list for adding new api:
- [X] code
- [X] tests
- [X] comments
- [X] example
- [X] readme

Please note that I had to make a decision regarding the `grid.extract` method tests. Changing this function to accept references broke all of the `papergrid>tests>extract` tests which I was not happy with. However, I do think that letting the `Grid` extract api be more obtuse to allow the `Table` extract api to be more fluent is a worthy trade.

Again any feedback would mean a lot. Cheers!


src/extract.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+impl<R, C> Extract<R, C>
+where
+    R: RangeBounds<usize>,
+    C: RangeBounds<usize>,
+{
+    pub fn new(rows: R, columns: C) -> Extract<R, C> {
+        Extract { rows, columns }
+    }
```
zhiburt, line 65: Maybe our current interface with `new` function is a bit ugly?

How do you feel about the next interface.

```rust
    // rows from 0 to max
    pub fn columns(columns: C) -> Extract<R, C>;
    // columns from 0 to max
    pub fn rows(rows: R) -> Extract<R, C>;
    pub fn segment(rows: R, columns: C) -> Extract<R, C>;
```

src/extract.rs
```diff
@@ -0,0 +1,76 @@
...
+///                    | : 2 : | : Hamburg : | : false : |\n\
+///                    +-------+-------------+-----------+\n\
+///                    | : 3 : |  : Brest :  | : true :  |\n\
+///                    +-------+-------------+-----------+\n");
+/// ```
+///
+pub struct Extract<R, C>
+where
+    R: RangeBounds<usize>,
+    C: RangeBounds<usize>,
```
zhiburt, line 52: As I understand you can just remove this block?

papergrid/src/lib.rs
```diff
@@ -396,7 +396,31 @@ impl Grid {
...
+    /// grid.extract(.., ..1)
+    /// +---+
+    /// |0-0|
+    /// +---+
+    /// |1-0|
+    /// +---+
+    /// |2-0|
+    /// +---+
+    /// ```
+    pub fn extract<R, C>(&self, rows: &R, columns: &C) -> Self
```
zhiburt, line 423: I may missed something but why these references are so  important?

```rust
rows: &R, columns: &C
```

zhiburt (changes requested):
Hi @IsaacCloos you've put quite an effort to this PR 💯 
Thank you for opening it.

Everything is very good.
Thank you for updating README.md and adding tests.

I've left a few comments.
Looking forward to hear your response

PS: If all my PRs were structured like yours probably someone even could consider me a good dev 😅

src/extract.rs
```diff
@@ -0,0 +1,76 @@
...
+    rows: R,
+    columns: C,
+}
+
+impl<R, C> Extract<R, C>
+where
+    R: RangeBounds<usize>,
+    C: RangeBounds<usize>,
+{
+    pub fn new(rows: R, columns: C) -> Extract<R, C> {
```
zhiburt, line 63: What do you think about creating a comment and mention what happens when ranges are bigger then the grid equal to it and equal to 0

README.md
```diff
@@ -341,6 +343,56 @@ Table::new(&data)
...
 ```
 
+### Extract
+
+You can `Extract` segments of a table to focus on a reduced number of rows and columns.
+
+```rust
+Table::new(&data)
+    .with(Extract::new(rows: 1..3, columns: 1..));
+```
```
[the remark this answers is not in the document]
IsaacCloos: I was copying the analyzer annotations as well in that snippet. I was going for clarity, but actually added confusion!
I will update this to remove those annotations.
To my knowledge it is impossible to inline write the input arguments (similar to C# for example)

papergrid/src/lib.rs
IsaacCloos: I could use some help on this issue. 
I made these changes to the `papergrid` extract api because I was unable to dereference the provided `RangeBounds` input from the `tabled` extract api.

I think it is super valuable to allow a user to write something like the following in the api of the exposed package.

```rust
Extract::new(.., 1..)
```

[TableOption change function](https://github.com/zhiburt/tabled/pull/87/files#diff-f3fcb4eeee8bd925542056f5ea55be80ef332b1c41fd746b30e0fc255813e8bfR68-R76)

I also see that you are suggesting to change the Extract api implementation to use functions over the New/TableOption api! That very well could address this issue entirely. Good idea I think

src/extract.rs
IsaacCloos: Absolutely! Thank you for the feedback. Really makes a huge difference 🙂

src/extract.rs
IsaacCloos: In my testing you are right. These can be removed. Will do

src/extract.rs
IsaacCloos: I will experiment shifting the implementation to be functional as you suggest. 

This will most likely correct the issue mentioned below with access to the `papergrid` extract api being locked behind a reference to `self` within the TableOption change function.

src/extract.rs
```diff
@@ -67,8 +67,21 @@ pub struct Extract<R, C> {
...
+            rows: rows,
+            columns: ..,
+        }
+    }
+    pub fn columns(columns: C) -> Extract<RangeFull, C> {
+        Extract { rows: .., columns }
+    }
+    pub fn segment(rows: R, columns: C) -> Extract<R, C> {
         Extract { rows, columns }
     }
```
zhiburt, line 86: I would add a doc comment here.

zhiburt, line 86: What do you think will it be better to separate the factions like this?

```rust
impl<R, C> Extract<R, C>
where
    R: RangeBounds<usize>,
    C: RangeBounds<usize>,
{
    pub fn segment(rows: R, columns: C) -> Self {
        Extract { rows, columns }
    }
}

impl<R> Extract<R, RangeFull>
where
    R: RangeBounds<usize>,
{
    pub fn rows(rows: R) -> Self {
        Extract {
            rows: rows,
            columns: ..,
        }
    }
}

impl<C> Extract<RangeFull, C>
where
    C: RangeBounds<usize>,
{
    pub fn columns(columns: C) -> Self {
        Extract { rows: .., columns }
    }
}
```

tests/extract_test.rs
```diff
@@ -125,7 +125,7 @@ fn extract_empty_test() {
         .with(Modify::new(Full).with(Alignment::left()))
         .with(Modify::new(Full).with(Indent::new(3, 1, 0, 0)))
         .with(Modify::new(Full).with(Format(|s| format!("[{}]", s))))
-        .with(Extract::new(1..1, 1..1))
+        .with(Extract::segment(1..1, 1..1))
         .to_string();
 
     assert_eq!(table, "");
```
zhiburt, line 131: Maybe to add 2 small tests for `Extract::columns` and `Extract::rows`

zhiburt (commented):
Looks good.
If you wish let me know and we can merge it as it is.

But I left a few comments.

papergrid/tests/extract.rs
```diff
@@ -18,7 +18,7 @@ fn extract_inside_test() {
 fn extract_left_test() {
     let grid = new_grid::<3, 3>();
 
-    let grid = grid.extract(.., ..1);
+    let grid = grid.extract(&(..), &(..1));
```
zhiburt, line 21: Can you remove these `&` references I am pretty sure it will work out?

IsaacCloos:
No merge yet! Busy with work. Will finish tonight with additional feedback. Thanks again for review, makes a huge difference 😄

src/extract.rs
IsaacCloos: The change has been made. Tests have been added/rewritten to accommodate this change.

Took a stab at adding doc references using this resource: [rustdocs](https://doc.rust-lang.org/rustdoc/linking-to-items-by-name.html#valid-links)

new apis:
- Extract::columns
- Extract::rows
- Extract::segment

src/extract.rs
IsaacCloos: individual docs have been added

IsaacCloos: I made this change as suggested. 

I'm new and would love to know what advantages this approach offers 😃

tests/extract_test.rs
IsaacCloos: `tabled>tests>extract.rs` has been extended by 6 additional tests for rows and columns

papergrid/src/lib.rs
IsaacCloos: Unfortunately, I have been unable to resolve this issue. 

The comment above explains the issue I'm still having.

IsaacCloos:
Commands I've been using to test:

`cargo test --test extract_test`
`cargo run --example extract`

papergrid/tests/extract.rs
zhiburt: Now I see the issue.
Look how I fixed it.
Maybe there's some better way but this is OK I think.

https://github.com/zhiburt/tabled/blob/0cb4d9dc0a4ee2c697f57783999fb2fd942abceb/src/extract.rs#L170-L178

zhiburt (approved):

zhiburt:
Thank you @IsaacCloos for your commitment.
And congrats with your first public pull request.

You did a great job.
Good documentation. 

Take care.

PS: Hopefully it won't be the last one :smile:

Issue #82, opened 2022-03-12
Add a `Margin` modifier


zhiburt:
> Margins
>
> Margins can be set outside the grid (top, bottom, left, right). Margins and the character used for rendering the margin can be set separately.
> 
> ```text
> vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
> >>>┌───────────┬───────────┐<<<<
> >>>│row 1 col 1│row 1 col 2│<<<<
> >>>├───────────┼───────────┤<<<<
> >>>│row 2 col 1│row 2 col 2│<<<<
> >>>└───────────┴───────────┘<<<<
> ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```

There's a list of steps necessary to take.

- A new `margin` field must be added to `papergrid::Grid`.
- `fmt::Display` of `papergrid::Grid` must be updated.
- `margin.rs` with `Margin` modifier must be created in `tabled`  which will set the table margin.

The possible implementation of `tabled::Margin`

```rust
struct Margin {
  top: Option<Indent>,
  bottom: Option<Indent>,
  left: Option<Indent>,
  right: Option<Indent>,
}

struct Indent {
  size: usize,
  symbol: char,
}
```

_______

Inspired by the same functionality in https://github.com/vdmeer/asciitable

kozmod:
I hope you don't mind that I'll get this one as the next task?

zhiburt:
No way I could mind against this :)
You're welcome.

zhiburt:
Hi @kozmod just realized 1 thing.

Maybe it's valuable to be able to set a `lead` side?

I mean what about the prior example I've provided.



> > ```
> > vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
> > >>>┌───────────┬───────────┐<<<<
> > >>>│row 1 col 1│row 1 col 2│<<<<
> > >>>├───────────┼───────────┤<<<<
> > >>>│row 2 col 1│row 2 col 2│<<<<
> > >>>└───────────┴───────────┘<<<<
> > ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> > ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> > ```

It could actually also look like the following.

```
>>>vvvvvvvvvvvvvvvvvvvvvvvvvvvvv
[...]

PR #77, opened 2022-03-08
Add TotalWidth support

zhiburt:

PR #89, opened 2022-03-19
[#82] add a `margin` modifier

kozmod:


papergrid/src/lib.rs
```diff
@@ -797,6 +811,7 @@ impl AlignmentVertical {
 pub struct Settings {
     text: Option<String>,
     padding: Option<Padding>,
+    margin: Option<Margin>,
```
zhiburt, line 814: I think `Margin` doesn't belong to `Settings` because settings are used to configure a segment of table, whether margin affects a table as a whole.

So it's better to create a `set_margin` method on `Grid` level I think.

src/margin.rs
```diff
@@ -0,0 +1,30 @@
+use crate::TableOption;
+use papergrid::{Grid, Indent};
+
+//todo add description
```
zhiburt, line 4: It would be cool to make it `todo` happen :)

Maybe a doc example would be also good.

src/margin.rs
```diff
@@ -0,0 +1,30 @@
...
+    pub fn new(left: usize, right: usize, top: usize, bottom: usize) -> Self {
+        Self(papergrid::Margin {
+            top: Indent::spaced(top),
+            bottom: Indent::spaced(bottom),
+            left: Indent::spaced(left),
+            right: Indent::spaced(right),
+        })
+    }
+
+    pub fn set_fill(mut self, left: char, right: char, top: char, bottom: char) -> Self {
```
zhiburt, line 17: I would also create a small doc comment here.

papergrid/src/lib.rs
zhiburt: As I see it's actually unused currently?

papergrid/src/lib.rs
```diff
@@ -919,13 +952,56 @@ impl std::fmt::Display for Grid {
...
+            writeln(f, |f| {
+                repeat_char(f, self.margin.bottom.fill, margin_row_widths)
+            })?
+        }
+
         Ok(())
     }
 }
 
+fn margin_line_widths(
```
zhiburt, line 969: It could be replaced by `Grid::total_width` but it's not merged yet (#77)

zhiburt (changes requested):
Hi @kozmod 

It looks good, thank you for your effort.

It's a pity that it's necessary to affect so much code in order to do left/right margin.
I was hoping that it will not be such a big change :sweat_smile:.
But you've done a lot of work :+1: . (including the fact that `papergrid` code quality is not outstanding I would imagine :smiling_face_with_tear: )
Please give me some time I'll try to think it through. (Maybe some refactoring in `master` will help)

And I've left a few more comments here.

papergrid/src/lib.rs
zhiburt: And seems like you've already created `set_margin`

kozmod: sorry, 
At first I added `Margin`  to `Settings`. After some refactoring, I forgot to remove `Margin` from `Settings` ....

papergrid/src/lib.rs
kozmod: 👌🏻
It's up to you, merge #77 or #82 first )

src/margin.rs
kozmod: added

papergrid/src/lib.rs
zhiburt: I did merged it a few minutes ago.
You can check that function out.

But don't worry too much I'll take a look at it myself if you will have no chance to.

zhiburt: Still not looked on improvements we could take to be more "adaptive" in papergrid rendering logic.

kozmod: 👌🏻

Issue #61, opened 2022-02-02
Add `AlignmentHorizontal::No` in order to not change the content and print it as it is 


zhiburt:
Currently it's not possible to have a cell which starts from spaces.

Example:

```rust
#[test]
fn table_value_starts_with_spaces() {
    let expected = "+------+\n\
                    | i32  |\n\
                    +------+\n\
                    |    1 |\n\
                    +------+\n";

    let table = Table::new(&["   1"]).to_string();

    assert_eq!(table, expected);
}
```

The test will fail because by default we use `HorizontalAlignment::Center`

zhiburt:
The change must be done here

https://github.com/zhiburt/tabled/blob/e54007acf74c81576140e2698baa661fa9236f29/papergrid/src/lib.rs#L685-L713

IsaacCloos:
If this ticket is still available, I'd like to take a stab at it 😄

zhiburt:
I think though the name is not ideal.
Maybe you have idea for a better name?

Also a note that one function will be necessary  to be created here.

https://github.com/zhiburt/tabled/blob/82ed6582f12ec939cf56333eba646e765b30b218/src/alignment.rs#L19-L59

zhiburt:
Actually correct me if I am wrong but `AlignmentHorizontal::No` will act as `AlignmentHorizontal::Left` just not using `trim()`.
Maybe maybe it's just worth to have a setting whether we allowed to use trim or not and by default doesn't do it?

https://github.com/zhiburt/tabled/blob/82ed6582f12ec939cf56333eba646e765b30b218/papergrid/src/lib.rs#L747-L750

https://github.com/zhiburt/tabled/blob/82ed6582f12ec939cf56333eba646e765b30b218/papergrid/src/lib.rs#L754-L756

What do you personally think about this `trim` we are using?
Maybe it's not what the person would expect while using an `Alignment`?

Here's a good example what `trim` does.

https://github.com/zhiburt/tabled/blob/82ed6582f12ec939cf56333eba646e765b30b218/papergrid/tests/styling.rs#L110-L132

If we would remove `trim`  the second column would be identical to the first.

zhiburt:
But maybe `AlignmentHorizontal::No` is OK except the name for sure :)

IsaacCloos:
What do you think of these name ideas:

- `AlignmentHorizontal::Inherited`
- `AlignmentHorizontal::Natural`
- `AlignmentHorizontal::Raw`
[...]

PR #99, opened 2022-03-24
Trim

IsaacCloos:
## Not found in this PR:
- disabling default trim
- updated unit tests

> These issues have been addressed in #98 

## Found in this PR:
- [X] code
- [X] documentation
- [X] example
- [X] readme
- [X] unit tests (for Trim)

Criticism encouraged! Happy to work at this until it's perfect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,7 +338,7 @@ Table::new(&data)
 
 - `FormatFrom` - Uses `Vec` elements as new content.
 - `FormatWithIndex` - Like `Format` but with `row` and `column` index in lambda.
-- `Trim` - Removes leading and trailing whitespace from range of cells 
+- `Trim` - Removes leading and trailing whitespace from content over specified range of cells
 
 ### Padding
 
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ An easy to use library for pretty printing tables of Rust `struct`s and `enum`s.
         * [Text in a top border](#Text-in-a-top-border)
     * [Alignment](#Alignment)
     * [Format](#Format)
+        * [Modifiers](#Modifiers)
     * [Padding](#Padding)
     * [Max width](#Max-width)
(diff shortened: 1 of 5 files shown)


zhiburt:
That's good enough.
We can't do any better I think.

I would only add trim as function to format.
And maybe actually removed `Trim` itself or make it private.

```rust
impl Format {
  fn trim() -> Trim {
    Trim
  }
}
```

Plus I would do check if there's any top/bottom empty lines. And skip them.

IsaacCloos:
> Plus I would do check if there's any top/bottom empty lines. And skip them.

You would like to retain new line escape characters? Current implementation removes all `White_Space` including `\n` and `\t`

zhiburt:
> You would like to retain new line escape characters? Current implementation removes all White_Space including \n and \t

Yes removal of empty lines. which present before actual content and after.

zhiburt:
It could be checked by something like this.

```rust
let trimmed_content = String::new();
let empty_lines_count = 0;
for line in content.lines() {
   let is_empty = line.trim().is_empty()
  if is_empty {
     empty_lines_count += 1;
     continue;
  }

     if empty_lines_count > 0 &&  trimmed_content.is_empty() {
         empty_lines_count = 0;
     }
      
     for _ in 0 .. empty_lines_count {
         trimmed_content.push('\n');
     }

      trimmed_content.push_str(line);
}
```

IsaacCloos:
here is the string I'm testing this on:

`"[\n\t\"foo\",\n\t{\n\t\t\"bar\":1,\n\t\t\"baz\": [\n\t\t\t2,\n\t\t\t3\n\t\t]\n\t}\n]"`

taken from comment on ticket #61

IsaacCloos:
PR #101 introduced new features to better dictate this setting to users

PR #135, opened 2022-05-16
tabled/ Improve rendering of show example

vojta7:
This change should improve the 'show' example. This example was flickering between table changes, it should be stable now.

Tested on Linux in Alacritty.

### DIFF
--- a/examples/show/src/main.rs
+++ b/examples/show/src/main.rs
@@ -9,13 +9,15 @@
 //!
 //! A convertation to gif is done via https://dstein64.github.io/gifcast/
 
-use std::{io::stdout, time::Duration};
+use std::{
+    io::{stdout, StdoutLock, Write},
+    time::Duration,
+};
 
 use crossterm::{
-    cursor::{self, Hide, MoveTo},
-    execute,
+    cursor, queue,
     style::Stylize,
-    terminal::{Clear, ClearType},
+    terminal::{Clear, ClearType, EnterAlternateScreen, LeaveAlternateScreen},
 };
 use tabled::{
     object::{Columns, Object, Rows},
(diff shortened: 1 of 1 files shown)


zhiburt:
Hi @vojta7 

Indeed you're right it is flickering.
I was thinking it is because of the timeouts I set :)

Thanks a ton for the improvement.

And sorry for the code which is inside this example....
Quite a lot of duplication there. 

So could you enlighten me what makes a fix?
Is it a `StdoutLock` or `EnterAlternateScreen`/`LeaveAlternateScreen`? (I bet on the first)

I'll try to merge it tomorrow if you don't mind.

zhiburt:
Thank you @vojta7 for your contribution.

Take care.

zhiburt:
By the way thank you for a style of a commit message.

vojta7:
> Thank you @vojta7 for your contribution.
> 
> Take care.

You are welcome and thank you to :smiley: 

> So could you enlighten me what makes a fix? Is it a `StdoutLock` or `EnterAlternateScreen`/`LeaveAlternateScreen`? (I bet on the first)

Actually, it was the call to the println! macro, it flushes output after every line (I think, that it uses Stdout wrapped in [LineWriter](https://doc.rust-lang.org/stable/std/io/struct.LineWriter.html) With stdout.write_all there is only a single call to write for each table. You can verify this with strace

```
strace -e write target/debug/show >/dev/null
```

Output with println! looks something like this:

```
write(1, "\33[2J\33[4;1H                    \342\224"..., 331) = 331
write(1, "                    \342\224\202 Release "..., 133) = 133
write(1, "                    \342\224\234\342\224\200\342\224\200\342\224\200"..., 321) = 321
write(1, "                    \342\224\202 May 25, "..., 135) = 135
write(1, "                    \342\224\224\342\224\200\342\224\200\342\224\200"..., 321) = 321
write(1, "\n", 1)                       = 1
write(1, "\33[4;1H", 6)                 = 6
```

and with write_all:

```
write(1, "\33[2J\33[4;1H", 10)          = 10
write(1, "                    \342\224\214\342\224\200\342\224\200\342\224\200"..., 1231) = 1231
write(1, "\33[4;1H", 6)                 = 6
```
The above output should be from the same "animation frame".

The rest of my commit isn't necessary to remove the flickering :smile:. There was a small fix with a hidden cursor and I used the AlternateScreen because I prefer it for tui with animations, so it leaves the terminal in its original state after the end.

Issue #134, opened 2022-05-16
Readme spelling mistake


nskobelevs:
Under Theme heading:

There are a list of ready to use styles. Each style can be **castomized**.

zhiburt:
Hi @nskobelevs 

Back in the day, teachers always considered me bad at languages :disappointed_relieved: 

Thank you for pointing this out.

Maybe you want to make these fixes your self?

zhiburt:
There's a few more from people on reddit.

> >There are a list of ready to use styles. Each style can be castomized.

> I suspect this is a typo.

ref: https://www.reddit.com/r/rust/comments/uqskpu/comment/i8tic40/?utm_source=share&utm_medium=web2x&context=3

> > For this porpouse you can use Border.
>
> Spotted this one too, not to pile on but to (hopefully) be helpful :)
> customized
> and
> purpose
> would be the fixes, respectively.

ref: https://www.reddit.com/r/rust/comments/uqskpu/comment/i8tx1lw/?utm_source=share&utm_medium=web2x&context=3

nskobelevs:
>Maybe you want to make these fixes your self?

Sure, I'll make a PR.

nskobelevs:
> MaxWidth sets a maximum width of an object. While tinkering content we don't forget about its color.

Can you clarify the second sentence please?

zhiburt:
> > MaxWidth sets a maximum width of an object. While tinkering content we don't forget about its color.
> 
> Can you clarify the second sentence please?

Meaning:

When we cut or increase a width of a cell we keep correct colors.
*at least suppose to :)

zhiburt:
closed by #138

PR #138, opened 2022-05-16
Fix grammar errors

nskobelevs:
This PR addresses #134 by fixing various grammar issues in the `README` as well as fixing a couple of code examples

One point of confusion:
>[While tinkering content we don't forget about its color.](https://github.com/zhiburt/tabled/blob/3dd43221ee9e46d3cad1d5f871d36434dd5af42a/README.md?plain=1#L398)

Is this saying that color is still handled correctly when using `MaxWidth` or `MinWidth`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,8 +395,8 @@ vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
 
 ### Max width
 
-`MaxWidth` sets a maximum width of an object.
-While tinkering content we don't forget about its color.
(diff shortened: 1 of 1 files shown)


zhiburt:
> Is this saying that color is still handled correctly when using MaxWidth or MinWidth? 

Correct

zhiburt:
Good,
thank you very much you did a lot more then were addressed in #134.

Take care.

close #134

PR #140, opened 2022-05-17
remove needless pass by value

danieleades:
pass by reference instead of by value where possible.

no measurable perf change on my machine.


zhiburt:
Hi @danieleades 

Honestly some of it were passed by value because of my consideration that the faster we `drop` the better.

Thank your for the PR.

zhiburt:
Hi @danieleades 

Just wanted to point out that now I think that generally pass a reference is better :sweat_smile: 

But I guess we can close the PR.

Thank you for the contribution.
Take care.

PR #144, opened 2022-05-20
Refactoring to reduce count of allocations

zhiburt:


zhiburt:
#148 covers this PR's logic.

PR #146, opened 2022-05-22
papergrid/ Create Borders structure

zhiburt:
Priviously we relied on HashMap only for borders.
But not everyone uses it.

Most of the time a constant Borders is enough.
So we not pay for what we don't use by allocations.


zhiburt:
moved to #148

PR #147, opened 2022-05-22
Use components in actions-rs/toolchain@v1

dmarcoux:
An additional step isn't needed. We can rely on `components`, a comma-separated list of the additional components to install.

Just something I've noticed when looking at the project.


zhiburt:
Hi @dmarcoux 

I've not known about this argument.

Apparently it will speed up a CI a bit.

Thank you for the contribution.

Issue #176, opened 2022-06-26
Custom hyperlinks cause wrong column borders


avsaase:
Hi,

Thanks for making this awesome crate. 

I'm making a command line tool that prints a table with some URLs. The URLs are too long to display in full so instead I want to show a custom text (see [here](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda) for more info on this). 

Here's a MWE using `tabled`:

```
use tabled::{object::Segment, Alignment, Modify, Table, Tabled};

#[derive(Debug, Tabled)]
struct MyStruct {
    name: String,
    #[tabled(display_with = "display_option_link")]
    link: Option<Link>,
    #[tabled(display_with = "display_option_link")]
    link2: Option<Link>,
}

#[derive(Debug, Tabled)]
struct Link {
    pub text: String,
    pub url: String,
}

impl Link {
    pub fn new(text: String, url: String) -> Self {
        Self { text, url }
    }
}

fn display_option_link(o: &Option<Link>) -> String {
    match o {
        Some(s) => format!("\u{1b}]8;;{}\u{1b}\\{}\u{1b}]8;;\u{1b}\\", s.url, s.text),
        None => format!(""),
    }
}

fn main() {
    let items = vec![
        MyStruct {
            name: "Google".to_string(),
            link: Some(Link::new(
                "click".to_string(),
                "https://www.google.com".to_string(),
            )),
            link2: Some(Link::new(
                "click".to_string(),
                "https://www.google.com".to_string(),
            )),
        },
        MyStruct {
            name: "Bing".to_string(),
            link: Some(Link::new(
                "click".to_string(),
                "http://www.bing.com".to_string(),
            )),
            link2: Some(Link::new(
                "click".to_string(),
                "http://www.bing.com".to_string(),
            )),
        },
    ];

    let table = Table::new(items)
        .with(Modify::new(Segment::all()).with(Alignment::left()))
        .to_string();
    println!("{}", table)
}
```

This prints
```
+--------+---------------------------------------+---------------------------------------+
| name   | link                                  | link2                                 |
+--------+---------------------------------------+---------------------------------------+
| Google | click | click |
+--------+---------------------------------------+---------------------------------------+
| Bing   | click    | click    |
+--------+---------------------------------------+---------------------------------------+
```

with the links clickable in a supported terminal emulator.

[...]

PR #201, opened 2022-08-08
 add support for terminal link and add test

kimlym:
The MR is mainly to add support for text as a terminal link.

Before: 
901659974798_ pic

After:
911659974826_ pic

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ derive = ["tabled_derive"]
 color = ["papergrid/color", "ansi-str"]
 
 [dependencies]
-papergrid = "0.5.1"
+papergrid = "0.5.2"
 unicode-width = "0.1.9"
 tabled_derive = { version = "0.4.0", optional = true }
 ansi-str = { version = "0.3.0", optional = true }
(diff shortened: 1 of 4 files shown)


zhiburt:
Hi @kimlym 

Thank you for openning the PR.
Personally I think `papergrid` has not very "user friendly" code base yet (WIP) so you did a good job.

Could you check this #176.
Just run your test with `cargo test --features color` will it work?
Maybe it will help.
I just wanna be sure it won't.

kimlym:
> Hi @kimlym
> 
> Thank you for openning the PR. Personally I think `papergrid` has not very "user friendly" code base yet (WIP) so you did a good job.
> 
> Could you check this #176. Just run your test with with no changes using `cargo test --features color` will it help? Maybe it will help. I just wanna be sure it won't.

Cool Thanks XD.

Sure thing, let me try and will paste the result here if necessary

kimlym:
Actually there would still be an issue after I apply Width::wrap, but that would be a different one

zhiburt:
> Actually there would still be an issue after I apply Width::wrap, but that would be a different one

You're actually right in both cases.
Doing wrap/truncate correctly would require a special logic to preserve the same link on the content.
Maybe we should open an issue to track it. (not sure)

zhiburt:
Hi @kimlym 
just wanted to let you know that it must be fixed on the latest version (`0.9.0`).

PS: the fix has some flaws but at least it shouldn't break a table.

Issue #203, opened 2022-08-10
Recognize terminal links when doing wrap/truncate


zhiburt:
As was pointed out in #201 currently when we do width manipulations we break links.
The idea to fix it to preserve links content.

If the logic will be quite complex it's better to be hidden by some feature flag.

dandavison:
Hi @zhiburt, I might have a go at this.

zhiburt:
Hey @dandavison sorry for the delay,

Thanks for the proposition.

I've implemented 1 wrap version already (there 2 version 1 which tries to keep words and another which doesn't).
So I don't know, maybe I'll continue? (It takes me a while.....)

I've seen your message about implementation proposal.
I stuck with regex. Maybe `vte` would be more efficient, but ... I don't know.

PS: `wrap.rs` is **quite** a messy place (duplication here and there etc.) I think you would not like the experience :) (joking)

PR #223, opened 2022-09-14
v0.8.0 Fix wrapped filenames with hyperlinks

dandavison:
Hi @zhiburt, this patch fixes nushell bug https://github.com/nushell/nushell/issues/6553 causing filenames in narrow tables to be incorrectly displayed, and fixes https://github.com/zhiburt/tabled/issues/203.  The patch was made against tabled v0.8.0 (the version of tabled that nushell is currently [using](https://github.com/nushell/nushell/blob/12a0fe39f73e88a2be97dce3d9cbe025b10e253d/crates/nu-table/Cargo.toml#L20)), but as you can see it's a tiny change, which can also be made against `src/features/width/wrap.rs` in your devel branch of tabled ([here's a branch](https://github.com/zhiburt/tabled/compare/master...dandavison:vte-ansi-iterator-hyperlinks) targeting your current master). I thought this PR targeting v0.8.0 might be more useful for getting the fix into nushell sooner. The crate it's using is [here](https://github.com/dandavison/vte-ansi-iterator). It has unit tests and has been well tested by users, running in delta for a couple of years now.

I've added one example: `cargo run --example hyperlink_keep_words` displays a table which is incorrect without the patch:

***On this branch***
```
 〉cargo run --example hyperlink_keep_words
.-----------------------------------------------.
| name                           | is_hyperlink |
| Debian                         | true         |
| Debian                         | false        |
| DebianDebianDebianDebianDebian | true         |
| DebianDebianDebianDebianDebian |              |
| DebianDebianDebianDebian       |              |
| DebianDebianDebianDebianDebian | false        |
| DebianDebianDebianDebianDebian |              |
| DebianDebianDebianDebian       |              |
'-----------------------------------------------'
```

***With the patch reverted***
```
 〉git revert --no-edit HEAD
 .-----------------------------------------------.
| name                           | is_hyperlink |
| De | true         |
| bian                      |              |
| Debian                         | false        |
| De | true         |
| bianDebianDebianDebianDebianDe |              |
| bianDebianDebianDebianDebianDe |              |
| bianDebianDebianDebian    |              |
| DebianDebianDebianDebianDebian | false        |
| DebianDebianDebianDebianDebian |              |
| DebianDebianDebianDebian       |              |
'-----------------------------------------------'
```

Currently it's simply removing the hyperlinks. I have the code to re-apply the hyperlinks to the wrapped segments, but I didn't want to disturb your `split_keeping_words` and `split` functions.


zhiburt:
Hi @dandavison

But I was thinking we shall preserve the links?
Or we can strip them?

I mean in case of wrapping where text is just a link with no other content, all lines would be links.
What do you think?

dandavison:
> But I was thinking we shall preserve the links?

Hi @zhiburt, good call! I agree. I didn't do that last night because I wanted to think about the API, but I'm going to have a go now. I'm thinking that the best way might be to return a prefix and a suffix, that can be reapplied to each fragment of text.

dandavison:
Hey @zhiburt, just a quick question: could you confirm that this is a bug in v0.8.0: https://github.com/zhiburt/tabled/commit/82e28c974bf65cd83e6a34f7ed165ddf23d04a72 I think you may have fixed it in master.

zhiburt:
> Hey @zhiburt, just a quick question: could you confirm that this is a bug in v0.8.0: https://github.com/zhiburt/tabled/commit/82e28c974bf65cd83e6a34f7ed165ddf23d04a72 I think you may have fixed it in master.

I think you're correct with a fix.
I think it's not fixed in the master.

https://github.com/zhiburt/tabled/blob/34e2d9376cab455f4272417b97f6cf2404ca4c2f/src/features/width/wrap.rs#L354

Do you working on this?
As I kind of stuck..... :disappointed: as I've noticed some issue in my logic (I was getting link indexes and strip the links via regex and then doing wrap while restoring links by offsets, but I forget about shift SGR gives (which we also "stripped" before)).

dandavison:
Thanks! Yes, I'm working on this. I'm almost ready to open a PR. Hopefully I can do it this evening. Here's my current diff: 

https://github.com/dandavison/tabled/compare/v0.8.0-vte-ansi-iterator-hyperlinks...v0.8.0-vte-ansi-iterator-hyperlinks-osc-partition

https://github.com/dandavison/vte-ansi-iterator/compare/main...osc-partition

dandavison:
Clearly we have to support stuff like this :)

```
printf '\e]8;;http://example.com\e\\\e[31mThis\e[0m \e[32mis\e[0m \e[33ma\e[0m \e[34mlink\e[0m\e]8;;\e\\\n'
```

image


(I wonder if Alacritty's even supporting that correctly? There's no link on "is a link")

examples/hyperlink_keep_words.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    let table = Table::from_iter(&data).with(Style::ascii_rounded()).with(
+        Segment::all()
+            .modify()
+            .with(Width::wrap(30).keep_words())
+            .with(Alignment::left()),
+    );
+
+    println!("{}", table);
+    println!("{}", multicolored_debian);
```
dandavison, line 72: This second print statement is just to check behavior while we're discussing this PR.

dandavison:
Hi @zhiburt, OK, this is very close I think.

```
cargo run --features=color --example hyperlink_keep_words
```

<table>
    <tbody>
        <tr>
            <td>image
</td>
        </tr>
    </tbody>
  </table>

So that is almost correct. There's just a small bug with the way the colors are starting too early, at the ends of the lines: image
 
The bug only occurs with `split_keeping_words`. Here it is using plain `split`:

<table>
    <tbody>
        <tr>
            <td>image
</td>
        </tr>
    </tbody>
  </table>



I'm guessing we should look for this bug in the previously existing code, but I might be wrong :)

Anyway, what do you think? Shall we fix nushell tables with an update to tabled including this? And if you're able to help find that but that would be awesome, but I'll look too.

dandavison:
nushell tables are looking correct. It looks like that bug doesn't affect unicolored file names:

<table>
    <tbody>
        <tr>
            <td>image
</td>
        </tr>
    </tbody>
  </table>

src/width.rs
```diff
@@ -1376,10 +1399,21 @@ fn split_string_at(text: &str, at: usize) -> (&str, &str, (usize, usize)) {
...
+    }
+
+    fn split(s: &str, width: usize, sep: &str) -> String {
+        super::split(s, width, "", "", sep)
+    }
+
+    fn chunks(s: &str, width: usize) -> Vec<String> {
+        super::chunks(s, width, "", "")
+    }
+
```
dandavison, line 1416: I can add some tests.

fdncred:
loving this team work - excited to see a fix in nushell, although a light colored terminal is questionable. 🤣

src/width.rs
```diff
@@ -1075,23 +1076,26 @@ where
 pub(crate) fn wrap_text(text: &str, width: usize, keep_words: bool) -> String {
     if width == 0 {
         String::new()
-    } else if keep_words {
-        split_keeping_words(text, width, "\n")
     } else {
-        split(text, width, "\n")
+        let (prefix, text, suffix) = osc_partition(text).unwrap_or(("", text, ""));
```
zhiburt, line 1080: I don't think it will work in general case.
I mean it will work in basic case.

But what if the string has 2 links?

```rust
    let data = [
        Distribution {
            name: format!(
                "{} 2 links in a string {}",
                format_osc8_hyperlink("https://www.debian.org/", "Debian"),
                format_osc8_hyperlink("https://www.wikipedia.org/", "Debian"),
            ),
            is_hyperlink: true,
        },
    ];
```

```
".--------------------------------.\n| name            | is_hyperlink |\n| \u{1b}]8;;https://www                | true         |\n| .debian.org/    |              |\n| \u{1b}\\Debian \u{1b}]8;; \u{1b}\\         |              |\n| 2 links in a    |              |\n| string  \u{1b}]8;;htt        |              |\n| ps://www.wikipe |              |\n| dia.org/\u{1b}\\Debian  |              |\n| \u{1b}]8;; \u{1b}\\                |              |\n'--------------------------------'"
```

src/width.rs
```diff
@@ -1139,6 +1145,7 @@ fn chunks(s: &str, width: usize) -> Vec<String> {
             continue;
         }
 
+        line.push_str(prefix);
```
zhiburt, line 1148: I assume you're implying that the string doesn't contains any other text?
But it may will.
And in such cases we would not like to make all lines being links.
Only the lines that are contains the actual link. (to be precise the content of the links)

```rust
    let data = [Distribution {
        name: format!(
            "here we get a debian link {}, right go and install it.",
            format_osc8_hyperlink("https://www.debian.org/", "Debian"),
        ),
        is_hyperlink: true,
    }];
```

```
".--------------------------------.\n| name            | is_hyperlink |\n| here we get a   | true         |\n| debian link  \u{1b}]8   |              |\n| ;;https://www.d |              |\n| ebian.org/      |              |\n| \u{1b}\\Debian \u{1b}]8;; \u{1b}\\         |              |\n| right go and    |              |\n| install it.     |              |\n'--------------------------------'"
```

zhiburt (changes requested):
Hi @dandavison sorry for the delay.

I think the PR accomplish a general case.
Where we have a string being considered to be a link.
But it may not be always the case, see my comments bellow.

Yes that's why it's not that an easy issue, and doing it with no allocations is even more complex.

What do you think @dandavison am I right here?

dandavison:
@zhiburt thanks very much for reviewing carefully and catching those two points. I've updated the branch to strip the hyperlinks in the cases that you identified.

I suggest that we leave it to future work to do the more sophisticated operations of reapplying parsed hyperlinks to the precise substrings that they should apply to. Meanwhile, nushell doesn't need that feature, and the current branch state will fix nushell. Does that sound OK to you?

src/width.rs
dandavison: Addressed by 5950c102c2146c7b0d6f5b6ad78831c8ed5086d6

src/width.rs
dandavison: Addressed by 5950c102c2146c7b0d6f5b6ad78831c8ed5086d6

zhiburt:
So you @dandavison return prefix/suffix only in case if 1 link is present and no additional text is there, otherwise you trim the links. Correct?

```rust
    let table = |text: &str| {
        let data = [Distribution {
            name: text.to_owned(),
            is_hyperlink: true,
        }];

        Table::from_iter(data)
            .with(
                Segment::all()
                    .modify()
                    .with(Width::wrap(30).keep_words())
                    .with(Alignment::left()),
            )
            .to_string()
    };

    let text = format!(
        "asd {} 2 links in a string {}",
        format_osc8_hyperlink("https://www.debian.org/", "Debian"),
        format_osc8_hyperlink("https://www.wikipedia.org/", "Debian"),
    );
    assert_eq!(
        table(&text),
        "+--------------------------------+--------------+\n\
         | name                           | is_hyperlink |\n\
         +--------------------------------+--------------+\n\
         | asd Debian 2 links in a string | true         |\n\
         | Debian                         |              |\n\
         +--------------------------------+--------------+"
    );

    let text = format!(
        "link: {}",
        format_osc8_hyperlink("https://www.debian.org/", "Debian"),
    );
    assert_eq!(
        table(&text),
        "+--------------------------------+--------------+\n\
         | name                           | is_hyperlink |\n\
         +--------------------------------+--------------+\n\
         | link: Debian                   | true         |\n\
         +--------------------------------+--------------+"
    );

    let text = format!(
        "{} :link",
        format_osc8_hyperlink("https://www.debian.org/", "Debian"),
    );
    assert_eq!(
        table(&text),
        "+--------------------------------+--------------+\n\
         | name                           | is_hyperlink |\n\
         +--------------------------------+--------------+\n\
         | Debian :link                   | true         |\n\
         +--------------------------------+--------------+"
    );
```

> I suggest that we leave it to future work to do the more sophisticated operations of reapplying parsed hyperlinks to the precise substrings that they should apply to. Meanwhile, nushell doesn't need that feature, and the current branch state will fix nushell. Does that sound OK to you?

I think you may be right.
Doing it like you did is good enough in most cases.

Could you add tests?
And could you hide the dependency after `color` feature?

Let me know if you don't have time.

dandavison:
> Could you add tests?
> And could you hide the dependency after color feature?

Will do!

dandavison:
@zhiburt I believe that this may be ready now. Your test suite was very helpful at finding some bugs. The function used from the `vte-ansi-iterator` library is [`strip_osc`](https://github.com/dandavison/vte-ansi-iterator/blob/0a29cceaf6b2f0fc2e2b77627dd198cf36742660/src/lib.rs#L35-L68):
```
/// Strip OSC codes from `s`. If `s` is a single OSC8 hyperlink, with no other text, then return
/// (s_with_all_hyperlinks_removed, Some(url)). If `s` does not meet this description, then return
/// (s_with_all_hyperlinks_removed, None). Any ANSI color sequences in `s` will be retained.
```

 Have fun reviewing the tests! You can test with

```
cargo run --features=color --example hyperlink_keep_words
```

 (If you push a branch forked off the `v0.8.0` tag then I'll retarget my PR at that branch.)

dandavison:
**Future work:**

If we want to, we can try to reduce allocations in subsequent PRs. E.g. it might be possible to return `Cow` from  [`strip_osc`](https://github.com/dandavison/vte-ansi-iterator/blob/0a29cceaf6b2f0fc2e2b77627dd198cf36742660/src/lib.rs#L35-L68) and sometimes not allocate.

It would also be possible to extend  [`strip_osc`](https://github.com/dandavison/vte-ansi-iterator/blob/0a29cceaf6b2f0fc2e2b77627dd198cf36742660/src/lib.rs#L35-L68) so that it returns all URLs encountered; currently it returns the first only. It would then be possible in theory to reapply the URLs to the correct substrings of wrapped text.

I need to open issues for some small bugs I found. Before I forget these are:

- [ ] the `first_word` true/false bug

Two bugs visible in this screenshot:

<table>
    <tbody>
        <tr>
            <td>image
</td>
        </tr>
    </tbody>
  </table>

- [ ] color starting to early, on the line before it should start
- [ ] tabled is inserting spaces between the non-word ANSI boundaries due to a `.push_str(' ')`

src/width.rs
```diff
@@ -1082,6 +1086,38 @@ pub(crate) fn wrap_text(text: &str, width: usize, keep_words: bool) -> String {
...
+pub(crate) fn wrap_text(text: &str, width: usize, keep_words: bool) -> String {
+    if width == 0 {
+        String::new()
+    } else {
+        let (text, url): (String, Option<String>) = strip_osc(text);
+        let (prefix, suffix) = if let Some(url) = url {
+            // https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda
+            let osc8 = "\x1b]8;;";
+            let st = "\x1b\\";
+            (format!("{osc8}{url}{st}"), format!("{osc8}{st}"))
```
dandavison, line 1099: I can remove the unnecessary allocation here.

zhiburt: I can imagine how it will look like :sweat_smile:

I guess we can leave it as it is (till better days) to not make *split_* function more complex.

zhiburt:
Hi there,
sorry for the delay (as always)

> color starting too early, on the line before it should start

I could be intentional could you show the string you've used?

> tabled is inserting spaces between the non-word ANSI boundaries due to a .push_str(' ')

True but where it may become an issue?

dandavison:
Hi @zhiburt, no problem!

> could you show the string you've used?

You can see the issue in the output of
```
cargo run --features=color --example hyperlink_keep_words
```

zhiburt:
> You can see it in the spaces between the word "Debian". (Even if you get rid of the hyperlink, there are still spaces inserted)

These spaces must be not colored right?

image

Thanks for pointing me it.
I'll work on it.

dandavison:
> These spaces must be not colored right?

Right:

1. The black and blue colored rectangles should not be there.
2. It should be `DebianDebianDebian`, not `Debian Debian Debian`

> I'll work on it.
Great!

I'm working on removing all usage of `ansi_term` from the `vte-ansi-tools` library that contains the `AnsiIterator`.

zhiburt:
About the issue with spaces.
I believe it's fixed on master.

image

dandavison:
> About the issue with spaces.
> I believe it was fixed on master earlier.

OK, great. I don't think it's immediately necessary for nushell, since nushell links are a single color. So I think it would be OK to create an initial release fixing nushell that doesn't contain the fix in master. (But if you know how to apply the fix to this branch then please feel free to push commits to my branch)

(By the way, I've now removed all usage of `ansi_term` from the `vte-ansi-tools` library.)

fdncred:
The only time (i think) that things are multiple colors of ansi is when using find. Here's an example. This is in the current main. I'm not sure this applies here though.
image

dandavison:
> The only time (i think) that things are multiple colors of ansi is when using find. Here's an example. This is in the current main. I'm not sure this applies here though.

Thank you @fdncred I think this is an excellent point, and very relevant. It gives a great real-world test that everything is working correctly.

On a first look, I sort of think there may be a bug in nushell: I think sometimes the way it is inserting the ANSI colors for the `find` result is breaking the hyperlink. That's without wrapping, and I wonder whether it's a bug in nushell rather than in tabled.

For example, try searching for a single letter like `ls | find t` in the nushell repo root. I am seeing that some of the files retain functioning symlinks and some so not. E.g. check out the one under my cursor in the screenshot below: the hyperlink is only rendered by on the front part of the word. (I don't have time to look into this any more now though!).


image

fdncred:
There could be a bug but I can't reproduce your findings on the latest main branch in Windows.
image

dandavison:
Ah, thanks. Looks like it's an Alacritty bug. Works on Wezterm. Do you see it on Alacritty? I can report it to Alacritty this evening.

fdncred:
I'm not sure how to enable links in alacritty. A quick search didn't reveal much. I enabled url: and laucher: but it doesn't seem to repond to it. so i'm not really sure what is wrong.

zhiburt:
I hope I resolved the conflicts.

I wonder if this is correct?

https://github.com/zhiburt/tabled/blob/42addd71b623c9a0047fd12b586699243c135367/tests/width_test.rs#L2468-L2483

Also the question do you plan to publish `vte-ansi-tools` on `crates.io`?

dandavison:
Great. What do you think about just copying `strip_osc` and `AnsiIterator` into a module in `tabled` for now, so that this fix can go ahead without waiting for the crate to be prepared? (documentation etc) I'd be happy with that as long as we can switch to a crate when one exists, and don't have much time right not for making the crate carefully.

zhiburt:
> What do you think about just copying strip_osc and AnsiIterator into a module in tabled for now, so that this fix can go ahead without waiting for the crate to be prepared? (documentation etc) I'd be happy with that as long as we can switch to a crate when one exists, and don't have much time right now for making the crate carefully.

That's actually sounds good.
I'll work on it.

But still is it correct?
I think a link must be there right?

https://github.com/zhiburt/tabled/blob/42addd71b623c9a0047fd12b586699243c135367/tests/width_test.rs#L2468-L2483

zhiburt (approved):

zhiburt:
> But still is it correct?
> I think a link must be there right?

It was correct,
When there's not only link we strip the link. Which is good enough as we agreed.

Thanks for your contribution once again.
@dandavison

dandavison:
Nice work finishing this off @zhiburt! I'm running your branch  https://github.com/nushell/nushell/pull/6286 in my local nushell to test it. I'm a bit busy the next two weeks but later on I'll look into getting that crate published, and I'd like to help out with benchmarking / optimizations if I get time.

zhiburt:
Hi @dandavison 

Wanted to notice you that;
I've spend a bit of time to switch to `vte` (basically your impl) in `ansitok`.

Some example of an usage.
https://gitlab.com/zhiburt/ansitok/-/blob/b963d7a83a8d928c78000a96e8bafe55a513205e/examples/parse_sgr.rs

I also would like to highlight 1 behavior.
I might be wrong but in an original logic.

Parsing this string `\x1b\x1b\x1b\x1b\x1b` would result in no yielded elements. Which in my mind must be 5 `ESC`?
I am pointing this out just in case it matter.

dandavison:
> can we try to keep the VTE parser functions used in tabled and ansitok fairly cleanly isolated, so that we have the option in the future of moving them into a VTE ANSI Parser crate

To be clear: my concrete hope here is that  delta and nushell and tabled / ansitok can share a single implementation of the VTE `ANSIIterator`, and share some of the associated utilities built using the `ANSIIterator`. I don't mind too much where code is located!

PR #225, opened 2022-09-17
Fix is_first_word bug

dandavison:
I believe this is correct but this is just a draft MR without tests, so that this (putative) bug isn't forgotten.

### DIFF
--- a/src/features/width/wrap.rs
+++ b/src/features/width/wrap.rs
@@ -351,7 +351,7 @@ fn split_keeping_words(s: &str, width: usize, sep: &str) -> String {
 
             line.push_str(word);
             line_width += word_width;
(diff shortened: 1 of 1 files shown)


zhiburt:
It is a issue indeed.

I hope you don't mind that I've created a different commit? (I mentioned you as a co-author, btw is it github or git feature?)

https://github.com/zhiburt/tabled/commit/b89bfbfd5adf4709c4ca757ffbd6e2b3ae375a32

https://github.com/zhiburt/tabled/blob/b89bfbfd5adf4709c4ca757ffbd6e2b3ae375a32/src/features/width/wrap.rs#L553-L556

Without the fix it was `"111 \n2341" `

Issue #126, opened 2022-05-10
Add `Theme` support


zhiburt:
The idea to have a text and border style which might be called `Theme`.

So it could look something like this.

![image](https://user-images.githubusercontent.com/20165848/167572825-9f76a5dd-d265-4989-b87c-fb9358f2b138.png)
Seen https://github.com/jedib0t/go-pretty/blob/main/table/images/table-StyleColoredBright.png

The implementation would require some color library which would require adding a new feature flag.

I'd like to note that the look like this is must be working already, but it must be set by the user by manual changes in cell content and cell borders.

zhiburt:
The theme could be responsible for a layout of a table, e.g #130 .

The implementation I have in mind is to have a `ThemeBuilder` which would be responsible for the layout and for text formatting.

If it will be merged some-when we probably need to put some important message to `Format` type that it should be used with caution.

zhiburt:
We could create `colorization` as a different structure.

```rust
table.with(Colorization::new(color1, color2).pattern(ColorizationPattern::Chess))
```

```rust
enum ColorizationPattern {
  Columns,
  Rows,
  Chess,
}
```

zhiburt:
Done by `Colorization`.
Though the idea of themes still valid; there's more things we could do with this for example as `Layout`.

PR #237, opened 2022-10-17
Add sharp border

wfxr:
Hi @zhiburt, thanks for maintaining this awesome project. Recently I used it in a new cli project. However I found a very common style is missing from tabled：

image

So I created this patch to:

- Add a common `sharp` style.
- Fix `rounded` style show case in readme.

BTW that would be nice to put a `.rustfmt.toml` in project root so that other contributors don't have trouble when editor feature like `auto format on save` is enabled.


zhiburt:
Hi @wfxr agree in both cases.

Could you create a small test (take a look at `tests/style_test.rs`)?

_________________________

I had an opinion that if you make things customazible enough these small changes will be easier to do on callers end.
But apparently we still fail to do so :disappointed_relieved: (we certainly do) 

For your style for example you could do

```rust
Style::modern()
    .off_horizontal()
    .horizontals(vec![HorizontalLine::new(1, Style::modern().get_horizontal())])
```

PS: This idea of pre-defined themes is something which yet not totally achieved (#126).

wfxr:
> Could you create a small test (take a look at `tests/style_test.rs`)? (Let me know if you'll have no chance to)

Done. These tests look very neat, cool!

> This idea of pre-defined themes is something which yet not totally achieved; Where for example modern style could have a list of "flavors" like "sharp", "rounded" etc.

I also like this idea. How about this apporach:
```rust
enum Corner {
    Sharp,     // ┌
    Rounded,   // ╭
    BoldSharp, // ┏
}

enum Grid {
    All,
    Header,
    Body,
}

fn modern(corner: Corner::Sharp, grid: Grid::Header) -> ... {
   ...
}
```

zhiburt:
> I also like this idea. How about this apporach:

It's OK,
But I guess these settings can be better wrapped to some "builder"?

Overall I thing that the level of abstraction of this thing must be higher than `Style`.
Cause not all styles have this `Corner/Grid` settings. (For example `psql` which is strict in its nature).

Also look at `markdown` for example. It doesn't have this also but it has something different.

```
| name | designed_by    | invented_year |
|------|----------------|---------------|
| C    | Dennis Ritchie | 1972          |
| Rust | Graydon Hoare  | 2010          |
| Go   | Rob Pike       | 2009          |
```

This uses no alignment.
But we could have 

1. left 

```
|:------|:----------------|:---------------|
```

2. right 

```
|------:|----------------:|---------------:|
```

3. center 

```
|:------:|:----------------:|:---------------:|
```

______________________

It's correct that we could have a separate builder for each style.
And `style::modern()` would return `ModernBuilder` which would have settings like `grid`, `corner` etc.

_______________

To conclude I just don't yet haven't came up with a good interface to all of this.

PR #249, opened 2022-10-31
tabled/ Add `Colorization` pattern

zhiburt:
It's kind of not that good.

ref: #126


zhiburt:
done 037062dfce2542a3992fe504fcb2eaca92126953

PR #254, opened 2022-11-30
Add a parallel support

zhiburt:


zhiburt:
Just to clarify; why it was not merged;

As I remember there was no win; at least from just `ryan` usage

joshtriplett:
Even for large tables, I think you're unlikely to get a win from parallelized table rendering.

PR #256, opened 2022-12-12
typo

fn-bruce:


zhiburt:
Hi @fn-bruce thanks for the change.

Take care

PR #259, opened 2023-01-10
feat: Support for BorderChar on Table level

assemmarwan:
Add support for `BorderChar` on Table Level as mentioned in the issue. 

ToDo List

- [ ] Add the feature to the codebase
- [ ] Add basic tests for the new feature

### DIFF
--- a/src/features/style/border_char.rs
+++ b/src/features/style/border_char.rs
@@ -1,6 +1,6 @@
 use papergrid::records::Records;
 
-use crate::{style::Offset, CellOption, Table};
+use crate::{style::Offset, CellOption, Table, TableOption};
 
 /// [`BorderChar`] sets a char to a specific location on a horizontal line.
 ///
@@ -71,6 +71,35 @@ where
                     table
                         .get_config_mut()
                         .override_vertical_border(pos, self.c, offset);
+                }
+            }
+        }
+    }
+}
+
+
+impl<R> TableOption<R> for BorderChar
+where
+        R: Records,
+{
+    fn change(&mut self, table: &mut Table<R>) {
+        let offset = self.offset.into();
+        let (count_rows, count_cols) = table.shape();
+
+        for row in 0..count_rows {
+            for col in 0..count_cols {
+                let pos = (row, col).into();
+                match self.horizontal {
+                    true => {
+                        table
+                            .get_config_mut()
+                            .override_horizontal_border(pos, self.c, offset);
+                    }
+                    false => {
+                        table
+                            .get_config_mut()
+                            .override_vertical_border(pos, self.c, offset);
+                    }
                 }
             }
         }


README.md
```diff
@@ -445,6 +445,34 @@ fn main() {
...
+        .to_string();
+
+    assert_eq!(
+        table,
+        "| 0     | 1     | 2 |\n\
+         |-------|:-----:|---|\n\
+         | Hello | World | ! |"
+    );
+}
+```
```
zhiburt, line 474: It seems to be a copy&paste from the above?
But it shall not have `Modify`, right?

tests/style_test.rs
```diff
@@ -2480,6 +2480,31 @@ test_table!(
     "y     2     y    2-0    y    2-1    y    2-2    y"
 );
 
+test_table!(
```
zhiburt, line 2483: I'd add a tests for last line, and a line in a middle.
(the same for vertical)

zhiburt (changes requested):
Hi @assemmarwan 

Thanks for contribution,

I think you misunderstood the issue a bit.
(As well as I did with the example in the issue :))

The idea is that you target a specific line in a table being not dependent on row/column.

For example you want to set a char at 3rd line horizontal line, it might be a 3rd row or it may be a 2nd or even 1st row.

________

It'd need to create a new public function/struct like

```rust
BorderChar::horizontal_exact(3, ':', Offset::End(0))
// or
ExactLine::horizontal(3, BorderChar::horizontal(':', Offset::End(0)))
```

I'd go with the 1st interface as it less general, though `ExactLine` sounds interesting for some cases but not exactly here.

______

The implementation is relatively simple as well.
You'd need to calculate the needed exact row/column based on relative offset and index.

1. iterate over rows/columns
2. Use `Table::get_width_ctrl`, `Table::get_height_ctrl` to get row/column height/width
3. Check if you belong to the corresponding row/column
4. If you do then iterate over row/column to find a cell by `Offset` in a similar way.

assemmarwan:
> Hi @assemmarwan
> 
> Thanks for contribution,
> 
> I think you misunderstood the issue a bit. (As well as I did with the example in the issue :))
> 
> The idea is that you target a specific line in a table being not dependent on row/column.
> 
> For example you want to set a char at 3rd horizontal line, it might be a 3rd row (with offset 0) or it may be a 2nd (with offset 1) or even 1st row (with offset 3).
> 
> It'd need to create a new public function/struct like
> 
> ```rust
> BorderChar::horizontal_exact(3, ':', Offset::End(0))
> // or
> ExactLine::horizontal(3, BorderChar::horizontal(':', Offset::End(0)))
> ```
> 
> I'd go with the 1st interface as it less general, though `ExactLine` sounds interesting for some cases but not exactly here.
> 
> The implementation is relatively simple as well. You'd need to calculate the needed exact row/column based on relative offset and index.
> 
> 1. iterate over rows/columns
> 2. Use `Table::get_width_ctrl`, `Table::get_height_ctrl` to get row/column height/width
> 3. Check if you belong to the corresponding row/column
> 4. If you do then iterate over row/column to find a cell by `Offset` in a similar way.

Ohhh, I think I got you. But if you don't mind share a sample table output when `BorderChar::horizontal_exact(3, ':', Offset::End(0))` is used to test against it. Appreciate it 😃

README.md
assemmarwan: My bad, I have fixed it now. The intention is to show the difference between Table level vs specific Row/Column.

assemmarwan:
@zhiburt Do I keep the misunderstood feature 😅 ? Or should I discard it?

zhiburt:
> If you don't mind share a sample table output when `BorderChar::horizontal_exact(3, ':', Offset::End(0))` is used to test against it.

original table

```
"+------+----------------+---------------+\n\
                | name | designed_by    | invented_year |\n\
                +------+----------------+---------------+\n\
                | C    | Dennis Ritchie | 1972          |\n\
                +------+----------------+---------------+\n\
                | Rust | Graydon Hoare  | 2010          |
+------+----------------+---------------+
| Go   | Rob Pike       | 2009          |
+------+----------------+---------------+
```

assemmarwan:
> > Do I keep the misunderstood feature 😅 ? Or should I discard it?
> 
> I guess it's better be discarded as we can do exactly this by.
> 
> ```rust
> Modify::new(Segment::all()).with(BorderChar ... )
> ```

Nice. I will revert my changes 👍🏻 

> 
>  > If you don't mind share a sample table output when `BorderChar::horizontal_exact(3, ':', Offset::End(0))` is used to test against it.
> 
> original table
> 
> ```
> +------+----------------+---------------+
> | name | designed_by    | invented_year |
> +------+----------------+---------------+
> | C    | Dennis Ritchie | 1972          |
> +------+----------------+---------------+
> ```
> 
> `BorderChar::horizontal_exact(2, '$', Offset::Start(0))`
> 
> ```
> +------+----------------+---------------+
> | name | designed_by    | invented_year |
> $------+----------------+---------------+
> | C    | Dennis Ritchie | 1972          |
> +------+----------------+---------------+
> ```
> 
> `BorderChar::horizontal_exact(2, '$', Offset::Start(10))`
> 
> ```
> +------+----------------+---------------+
> | name | designed_by    | invented_year |
> +------+--$-------------+---------------+
> | C    | Dennis Ritchie | 1972          |
> +------+----------------+---------------+
> ```
> 
> `BorderChar::horizontal_exact(3, '$', Offset::Start(2))`
> 
> ```
> +------+----------------+---------------+
> | name | designed_by    | invented_year |
> +------+----------------+---------------+
> | C    | Dennis Ritchie $ 1972          |
> +------+----------------+---------------+
> ```

Thanks for the explanation 😃 

> 
> To be honest I got stuck thinking about:
> 
> 1. whether we shall allow indexing of intersections here (see 1st example)

For the sake of consistency with the current implementation, I would say not to allow it. But if we were to allow it, how would we achieve that with current implementation? If you would point me to how it's done please.


> 2. whether we shall index only horizontal lines or all lines (see 3rd example)

I think to remain consistent with the original `BorderChar` implementation, we index horizontal lines only. Example 3 can be achieved using `vertical_exact` function.

```rust
BorderChar::vertical_exact(2, '$', Offset::Start(3))
```

zhiburt:
> For the sake of consistency with the current implementation, I would say not to allow it.
>
> I think to remain consistent with the original BorderChar implementation, we index horizontal lines only.

Then actually we could achieve it by `Modify::new(Rows::new(your_row_goes_here)).with(BorderChar('$', Offset::Start(7)))`
I mean it then makes no scenes right? 

So I would try to do it my way?

> But if we were to allow it, how would we achieve that with current implementation? Because all offsets start without the external border (intersection) always. If you would point me to how it's done please.

You'd need to call `cfg.get_border()` to get borders of a given cell and then modify the `*_corner` field of it after which call `cfg.set_border()` to update it's intersection symbol.

So yes I tend to believe it's possible to do.
The algorithm basically stays the same.

> 1. iterate over rows/columns
> 2. Use Table::get_width_ctrl, Table::get_height_ctrl to get row/column height/width
> 3. Check if you belong to the corresponding row/column
> 4. If you do then iterate over row/column to find a cell by Offset in a similar way.
>     1. (Consider intersection as 1 step)
>

assemmarwan:
Alright, so I'll work on it such that:

- All horizontal lines are allowed
- Intersection is allowed

assemmarwan:
I have noticed that the functions `Table::get_width_ctrl` & `Table::get_height_ctrl` are not public. 

https://github.com/zhiburt/tabled/blob/64d68081580e05e7efe25945daa431a691cf2cda/src/table.rs#L293-L313


What do you recommend that I use to determine the target position of the character?

zhiburt:
> I have noticed that the functions `Table::get_width_ctrl` & `Table::get_height_ctrl` are not public grimacing .

I've forgot about it. (it actually a subject to change in the future).
You can use `get_table_widths` and `get_table_total_height2` in the meantime.

> What do you recommend that I use to determine the target position of the character?

What do you mean?
I think this must work?

> * iterate over rows/columns
> * Use Table::get_width_ctrl, Table::get_height_ctrl to get row/column height/width
> * Check if you belong to the corresponding row/column
> * If you do then iterate over row/column to find a cell by Offset in a similar way.
>   (Considering intersection as 1 step)
>   
>   1. while i < offset:
>   2. if i == offset => then it's intersection symbol
>   3. for i = i; i <= row_length; i++
>   4. if i == offset => then it's offseted char

assemmarwan:
@zhiburt Would you please assist about how to tell whether a cell has borders. This is assuming I have the border. 

is this way valid?

```rust
let mut border = table.get_config_mut()
                .get_border((0, 0), (count_rows, count_cols));

let has_border = border.top.is_some();

```

zhiburt:
> is this way valid?

Yes

```
                  _______   < ---- top border
                 |       | 
                 |  cell |  < ---- right border
left border ---> |       |
                  _______   < ---- bottom border
```

zhiburt:
Accidentally I needed to experiment with this;

And I've got some work in progress patchwork.

I'd say it's quite bad from performance perspective but it seems to work (I didn't test it well, just have a few runs).

```rust
impl<R> tabled::TableOption<R> for SetHorizontalChar
where
    R: Records,
{
    fn change(&mut self, table: &mut tabled::Table<R>) {
        let shape = table.shape();

        let is_last_line = self.line == (shape.0 * 2);
        let mut row = self.line;
        if is_last_line {
            row = self.line - 1;
        }

        let mut evaluator = tabled::papergrid::width::WidthEstimator::default();
        evaluator.estimate(table.get_records(), table.get_config());
        let widths: Vec<_> = evaluator.into();

        let mut i = 0;
        for column in 0..shape.1 {
            let has_vertical = table.get_config().has_vertical(column, shape.1);

            if has_vertical {
                if self.position == i {
                    let mut border = table.get_config().get_border((row, column), shape);
                    if is_last_line {
                        border.left_bottom_corner = Some(self.c1);
                    } else {
                        border.left_top_corner = Some(self.c1);
                    }

                    table.get_config_mut().set_border((row, column), border);

                    return;
                }

                i += 1;
            }

            let width = widths[column];

            if self.position < i + width {
                let offset = self.position + 1 - i;
                // let offset = width - offset;

                table.get_config_mut().override_horizontal_border(
                    (self.line, column),
                    self.c2,
                    tabled::papergrid::Offset::Begin(offset),
                );

                return;
            }

            i += width;
        }

        let has_vertical = table.get_config().has_vertical(shape.1, shape.1);
        if self.position == i && has_vertical {
            let mut border = table.get_config().get_border((row, shape.1), shape);
            if is_last_line {
                border.left_bottom_corner = Some(self.c1);
            } else {
                border.left_top_corner = Some(self.c1);
            }

            table.get_config_mut().set_border((row, shape.1), border);
        }
    }
}
```

PR #261, opened 2023-01-15
fix json_to_table test

danieleades:
this test has the wrong type in the constructor

i guess the CI is not covering this feature?
looks like the `check` target is not checking `--all-targets` or the whole workspace (`--all`)


danieleades:
i've updated the check target while adding the MSRV in #264

zhiburt:
> i guess the CI is not covering this feature?

You seem to be right,

Thanks for the fix.

Take care

PR #262, opened 2023-01-15
use inline format args (clippy::uninlined_format_args)

danieleades:


zhiburt:
Hi @danieleades 

Sorry for such a DELAY ....
I am actually not sure why I didn't answered earlier....

So I actually quite against this change as it will require some recent compiler version, which I tend to believe not good?
As far as I understand it won't compile on some versions right?

What do you think?

PS: Yes it's tests so maybe it doesn't matter a lot.

zhiburt:
Welllllll

It seems to be a default now....

https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args

danieleades:
> Welllllll
> 
> It seems to be a default now....
> 
> https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args

yes and no. Whether this lint gets applied or not depends on the configured MSRV. So it will depend on the decision in #264

zhiburt (approved):
Seems good;

Sorry for delay

PR #264, opened 2023-01-15
configure MSRV

danieleades:
- adds the MSRV to Cargo.toml
- adds a workflow target to check the MSRV

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,16 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.rust }}
-      - run: cargo check --no-default-features --features=${{ matrix.features }}
+      - run: cargo check --all-targets --all --no-default-features --features=${{ matrix.features }}
+
+  msrv:
+    name: MSRV
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: dtolnay/rust-toolchain@1.58.1
+      # check only the root crate for MSRV
+      - run: cargo build --all-features
 
   fmt:
     name: Rustfmt
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,8 @@ members = [
     "table_to_html",
     "benches",
 ]
+[workspace.package]
+rust-version = "1.58.1"
 
 [[example]]
 name = "colored_borders"


zhiburt:
I think having MSRV would be great, but I think maybe 1.59 is too optimistic version?
A bit explained in #262.

I guess finding the lowest version we can run tabled and sticking to it would be ideal?
Or maybe the version with `matches!` (I like this one).

zhiburt:
I've run `cargo msrv` and it shows that it's `1.56.1`

danieleades:
> I've run `cargo msrv` and it shows that it's `1.56.1`

What command did you use? Did you check all targets?

If not, are you comfortable having separate MSRVs for deployment and dev?

zhiburt:
Switching edition to 2018 from 2021 on some crates makes it `1.53.0` (only for dev target exactly).

> If not, are you comfortable having separate MSRVs for deployment and dev?

Personally I feel like it could be all right?
Or do you think it's a bad option?

______

Just found that `Copy` is allowed in `const` context only after `1.61` :(
Maybe sticking to very low version is not that good :thinking: 

PS: Sorry for such a delays; I was keeping the issue in mind just was not sure about the proper address of it. But surely it must be decided.

danieleades:
@zhiburt i've updated this PR to not check all targets when checking MSRV.

It's important to note that the MSRV is currently set in the workspace Cargo.toml file, but that workspace inheritance is not actually supported until 1.64.0. When the MSRV is bumped above 1.64.0 this should be revisited. see https://betterprogramming.pub/workspace-inheritance-in-rust-65d0bb8f9424

finally, I want to stress that the point of this PR is not to make a decision on a permanent, static MSRV. It is simply to provide a check such that the MSRV doesn't get bumped *accidentally* without an accompanying version bump. It can still be bumped freely in order to update dependencies, use newer syntax, etc.

danieleades:
after some experimentation, i've found that the MSRV when *building* the root crate (only) is 1.58.1. I believe this is the MSRV that downstream users will actually be exposed to when they try and add this library as a dependency while using an older compiler.

you can check this yourself with

```sh
cargo msrv -- cargo build --all-features
```

danieleades:
I notice there's a lot of 'fixup' commits from my last PR that have been merged into master without being squashed. Would you prefer me to squash these commits in the PR before you merge? Or do you want to squash merge them?

zhiburt:
> I notice there's a lot of 'fixup' commits from my last PR that have been merged into master without being squashed. Would you prefer me to squash these commits in the PR before you merge? Or do you want to squash merge them?

Yes.....
Someone messed it up ...

> after some experimentation, i've found that the MSRV when building the root crate (only) is 1.58.1

likely so.

_________________________________

Just to clarify I did not ignore the PR just kind of was thinking about it.

As I said we can refactor code a bit and will able to make it  `~1.53.0`.
But is it worth it?

So what do you suggest stick to `1.58`?

zhiburt:
Agree with your CI statement.

Can you update it to use `dtolnay/rust-toolchain`?

zhiburt:
Thanks a lot;

And sorry that it took that long...

Take care.

PR #266, opened 2023-01-15
use unwrap_or_default

danieleades:


src/features/measurement.rs
```diff
@@ -65,9 +65,9 @@ impl Measurement<Width> for Min {
     {
         let ctrl = CfgWidthFunction::from_cfg(cfg);
         grid_widths(&records, &ctrl)
-            .map(|r| r.min().unwrap_or(0))
+            .flatten()
```
zhiburt, line 68: Are you sure it is correct?

zhiburt: Applies to all the .flatten()

json_to_table/src/table.rs
```diff
@@ -335,14 +335,14 @@ mod json_to_table {
...
-                table.with(Width::increase(width.unwrap_or(0)));
+                table.with(Width::increase(width.unwrap_or_default()));
                 table.with(SetBottomChars(
                     used_splits,
                     table
                         .get_config()
                         .get_borders()
                         .top_intersection
-                        .unwrap_or(' '),
+                        .unwrap_or_default(),
```
zhiburt, line 345: Is ' ' actually a default?
I was always thinking it's '\0'

Interesting if so.

src/features/measurement.rs
danieleades: yep. `grid_widths` returns an iterator of iterators. `flatten` flattens the iterators into a single iterator of all values

json_to_table/src/table.rs
danieleades: I stand corrected- the default char is `\x00`.

I'll revert this change

src/features/measurement.rs
danieleades: actually, this one is wrong but the rest are right. This particular case is taking the max of the min, rather than the global max

zhiburt: actually, you're right (At first glance I was thinking all of wrong)

zhiburt (approved):
Thanks.

Sorry for the delay.

Take care.

PR #271, opened 2023-01-29
[#269] Add a table consistency property based test

### LINKED ISSUE #269: Add a table consistency property based test
It's need to

1. Add a test (better a few tests) which takes some data, and some settings (padding, row spans, columns spans, etc.).
2. Build a table with the settings and the data given.
3. And then check whether the table string contains artifacts (meaning broken).

    It can be done by chaning a style's left/right corner to some character (for example `%`),
    and checking if the character always stays right/left most.

_________________

We already have 1 similar test

https://github.com/zhiburt/tabled/blob/294118fa5a81f14c981856d560256c74f8011716/tests/builder_test.rs#L741-L758

__________________

To give you an idea this is how a broken table may look like

image

arunma:
Fixes #269


zhiburt (commented):
Hi @arunma,

So you almost there.

_____________________

I'd start from a general test.

Having a fixtures like this

```rust
struct DataFixture {
   cells: Vec<Vec<Lines>>
   theme: ThemeFixture,
}

type Lines = Vec<String>;

enum ThemeFixture {
  Ascii,
  Rounded,
  Empty,
  ...
}
```

Then,

1. join Lines into a string to build a table
2. build a table with a given theme
2. verify that all lines have the same width.

_________________________________________

In a second test you'd need to add a `row_span:usize` && `column_span:usize` fields, and use them when building a table.

You can also just compare the width of lines of resulting table to verify the results.


PS: Yes such assertions are not ideal but they good enough and we can start here.

tests/builder_test.rs
```diff
@@ -756,3 +766,115 @@ fn qc_table_is_consistent(data: Vec<Vec<isize>>) -> bool {
...
+    thirteen: bool,
+}
+
+impl Arbitrary for DataFixture {
+    fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+        let mut rng = StdRng::seed_from_u64(u64::arbitrary(g));
+        DataFixture {
+            one: Alphanumeric
+                .sample_string(&mut rng, u8::arbitrary(g) as usize)
+                .to_string(),
```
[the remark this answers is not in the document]
arunma: Indeed it does. Thanks !

tests/span_tests_qc.rs
```diff
@@ -0,0 +1,233 @@
...
+            RestructuredText,
+            AsciiRounded,
+        ])
+        .unwrap()
+        .to_owned()
+    }
+}
+
+#[test]
+#[ignore = "Test fails"]
```
zhiburt, line 63: So it fails?

zhiburt: Interesting if so.

tests/span_tests_qc.rs
```diff
@@ -0,0 +1,233 @@
...
+        tabled::papergrid::util::string_width(table_lines.first().unwrap_or(&"".to_owned()));
+
+    assert!(table_lines
+        .iter()
+        .all(|l| tabled::papergrid::util::string_width(l) == line_width));
+}
+
+#[quickcheck_macros::quickcheck]
+#[ignore = "Test fails"]
+fn qc_table_is_consistent_with_spans(table_structure: TableStructure) {
```
zhiburt, line 151: Maybe having 3 tests would be better?

1. which you have.
2. only with row spans
3. only with column spans

What do you think?

zhiburt (approved):
:100: 

Looks good (beside some minor things).

See it started to work :disappointed: 

https://github.com/zhiburt/tabled/actions/runs/4084893972/jobs/7343369098#step:4:318

Don't worry about my comments bellow.

Thanks for your contribution.
Take care.

_____________

Sorry that it took me these days to respond.

tests/builder_test.rs
```diff
@@ -756,3 +758,127 @@ fn qc_table_is_consistent(data: Vec<Vec<isize>>) -> bool {
...
+        Dots => {
+            table.with(Style::dots());
+        }
+        RestructuredText => {
+            table.with(Style::re_structured_text());
+        }
+        AsciiRounded => {
+            table.with(Style::ascii_rounded());
+        }
+    }
```
zhiburt, line 874: (Just a note) I could be moved to a function.

tests/span_tests_qc.rs
```diff
@@ -0,0 +1,233 @@
+use quickcheck::Arbitrary;
+use tabled::{builder::Builder, object::Cell, Modify, Span, Style};
+
+#[derive(Clone, Debug)]
+struct TableStructure {
+    pub rows: Vec<Line>,
+    pub theme: ThemeFixture,
+    pub row_span: Vec<u8>,
+    pub col_span: Vec<u8>,
+}
```
zhiburt, line 10: (Just a note);

See that you could use the same structure in both files.
I guess we could move it to a submodule and reuse it.

I was actually thinking about it for a while, to create a different   cargo crate for qc_tests.
Because they require additinal dependencies which increases build time.

tests/span_tests_qc.rs
```diff
@@ -0,0 +1,233 @@
...
+
+type Line = Vec<u32>;
```
zhiburt, line 12: (a note)

Maybe it's even better then `Vec<String>`.
From speed perspective at least but I don't know.

tests/builder_test.rs
```diff
@@ -756,3 +758,127 @@ fn qc_table_is_consistent(data: Vec<Vec<isize>>) -> bool {
...
+
+    let output = table.to_string();
+
+    let table_lines: Vec<String> = output.lines().map(|l| l.into()).collect();
+    let line_width =
+        tabled::papergrid::util::string_width(table_lines.first().unwrap_or(&"".to_owned()));
+    assert!(table_lines
+        .iter()
+        .all(|l| tabled::papergrid::util::string_width(l) == line_width));
```
zhiburt, line 883: (note)

Could be shortened to

```rust
let first_line =  output.lines().next().unwrap_or("");
assert_eq!(util::string_width_multiline(output), util::string_width(first_line));
```

zhiburt (commented):
just notes; you can ignore it;

PR #280, opened 2023-02-09
deprecates most usages of 'actions-rs'

danieleades:
'actions-rs' is unmaintained. This PR removes *most* of the usages from CI


zhiburt:
Thanks for the update.

zhiburt:
I guess we can also remove traupaline dependency?

https://github.com/xd009642/tarpaulin#github-actions

PR #281, opened 2023-02-09
add dependabot config

danieleades:
enables dependabot for both rust and github action dependencies.

This should not be merged until after #264, otherwise dependency updates may result in unintended MSRV bumps that won't be caught by CI

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: cargo
+  directory: "/"
+  schedule:
+    interval: daily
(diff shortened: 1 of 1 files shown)


danieleades:
@zhiburt this can be merged as soon as #264 is merged

danieleades:
just had a thought, not sure if dependabot will open PRs for bumping the MSRV toolchain action. If so, will need to add an 'ignore' for that action, since we don't want that updated by dependabot

danieleades:
@zhiburt this should be good to go now!

zhiburt:
Thanks for your commitment;

Take care.

PR #283, opened 2023-02-16
Symlink license file to each crate

michel-slm:
This allow each crate packaging to bundle the LICENSE file, as needed when distributing MIT license code


zhiburt:
Hi @michel-slm 

Thanks for the commit.

I have 2 questions in my minds in this domain.
If you don't mind I'd like to hear your thoughts.

1. I guess the MIT license notice must be present in each file too?
2. I've seen some projects with dual license MIT+Apache maybe we shall do it too? (I am not versed in all these licenses for me unilinse looks OK too :)) 

Take care.

michel-slm:
Hi @zhiburt !
- I don't think the notice needs to be present in each file, at least Fedora does not normally enforce it. Some other licenses like GPL on the other hand do require license headers on each file, though that often does not happen either. For Rust the natural unit of distribution is a crate, so having it in the crate is necessary
- yeah, a lot of Rust projects are dual MIT/Apache licensed because Rust itself is. Up to you; MIT is much more permissive than Apache so you can mix the two anyway. Apache only will be bad, as then the code can't be combined with GPLv2 (but can with v3)

Thanks for merging! Sorry for the late reply

PR #289, opened 2023-02-17
bump 'clap' dependency

danieleades:


zhiburt (approved):

zhiburt:
Thanks; for the bump

(Have you noticed like your CI catched MSRV?:)) (not sure in which PR)

danieleades:
If you recall, the CI is only checking the MSRV of the root crate

zhiburt:
Not sure if you noticed but I've changed the project structure so it's not any more. (subject to change cause noticed that build time got worth)
It is makes it so for all the crates in a workspace I believe.

PR #294, opened 2023-02-28
example instruction corrections

IsaacCloos:
Several minor changes to fix inconsistencies and incorrect top-level comments for examples. Many of these corrections fix bugs that prevented some examples from being run from the root of the crate.

# Criticism welcome 😄 
I'd be happy to change anything about this work to meet your standards.


IsaacCloos:
Please note that examples [[grid_colors.rs](https://github.com/zhiburt/tabled/blob/master/tabled/examples/grid_colors.rs), [highlight_color.rs](https://github.com/zhiburt/tabled/blob/master/tabled/examples/highlight_color.rs), [hyperlink.rs](https://github.com/zhiburt/tabled/blob/master/tabled/examples/hyperlink.rs)] are still omitted from the readme because they require a hosted image solution like [color.rs](https://github.com/zhiburt/tabled/blob/master/tabled/examples/color.rs)

https://github.com/zhiburt/tabled/blob/06e5feaa10308e2c2913109f190307619b118905/tabled/examples/README.md?plain=1#L120-L125

zhiburt:
Hi @IsaacCloos 

Yes it got outdated, as well as README.md. 
Thanks for the fix.

I wonder what do you think about putting the results of examples right to the head comment of the example?
We have  the results in examples/README.md but I wonder if it would be any useful to have it right there.

IsaacCloos:
> I wonder what do you think about putting the results of examples right to the head comment of the example? We have the results in examples/README.md but I wonder if it would be any useful to have it right there.

I think that would be a great idea! Cleaner for the user, easier for contributors.

PR #297, opened 2023-03-05
update table shape with vec_records insert_row function

IsaacCloos:
This change will keep `records.count_rows()` up-to-date with the tables shape after row insertion.

# Criticism Welcome!
This bug was impacting my work on a new Split setting. If I've missing something obvious please let me know 😄


zhiburt:
Good catch: I am not sure how it was missed. ....

Issue #290, opened 2023-02-20
Add `Split` settings.


zhiburt:
Imagine you have an alphabet.
It's easy to create it as a 1 line table.
But how to split it by index? (it's possible but yet not that easy)

```rust
use std::iter::FromIterator;
use tabled::Table;

fn main() {
    let table = Table::from_iter(['a'..='z']);
    println!("{table}");
}
```

```
+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
| a | b | c | d | e | f | g | h | i | j | k | l | m | n | o | p | q | r | s | t | u | v | w | x | y | z |
+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
```

So the idea is to create a new settings to split table by index (`Split::column`)

```rust
use std::iter::FromIterator;
use tabled::Table;

fn main() {
    let mut table = Table::from_iter(['a'..='z']);
    table.with(Split::column(4));
    println!("{table}");
}
```

```
+---+---+---+---+
| a | b | c | d |
+---+---+---+---+
| e | f | g | h |
+---+---+---+---+
| i | j | k | l |
+---+---+---+---+
| m | n | o | p |
+---+---+---+---+
| q | r | s | t |
+---+---+---+---+
| u | v | w | x |
+---+---+---+---+
| y | z |   |   |
+---+---+---+---+
```

IsaacCloos:
# ✋🏻 
If something isn't already in the works I'd love to draft an implementation of this! I can cook something up by the end of this week, and if you have any extra details/requests please let me know 😄

zhiburt:
Hi @IsaacCloos
You're welcome.

> if you have any extra details/requests please let me know

I was thinking maybe we could also add a `Builder::uniform` method to do that in place.
At least from perf. reasons it will be better than building a table and calling `Split`.

```rust
let data = vec![1, 2, 3, 4, 5];
let count_columns = 2;

let table = Builder::uniform(data, count_columns).build();
```

IsaacCloos:
@zhiburt 

What are your thoughts on `splitting` multi-row tables? I can make the `TableOption` `change` function resolve in a few ways:

1. cut on the index and shift all columns beyond to new rows starting below the last current row. This would, in a sense, treat the entire table as a single row while splitting. 
```
                                   +---+---+
                                   | a | b |
[...]

PR #298, opened 2023-03-07
add new function `insert_column` to Resizable trait

IsaacCloos:
I have a draft of Split ready to put up, but I needed to add this function to achieve all of the options discussed in https://github.com/zhiburt/tabled/issues/290. I figured it would be safer to split this dependent work into it's own PR so we can workshop any necessary changes you'd like to see before this goes in.

|          | Append | Zip    |
|:--------:|:------:|:------:|
| column(usize)   |   ✅    |   ✅    |
| row(usize)      |   ✅    | Requires changes in this PR |

---

P.S. I'm getting a yellow squiggly that says 
>missing documentation for an associated function

I'm not sure if that's referring to generated documentation online? I've added a doc comment above the function like insert_row().
https://github.com/zhiburt/tabled/blob/2319a19b0c80aa504476d68f2146211326bf43ed/tabled/src/records/mod.rs#L104-L105
I'm not sure what the difference is. I'm still learning about how Rust works 😅

### DIFF
--- a/tabled/src/records/mod.rs
+++ b/tabled/src/records/mod.rs
@@ -101,8 +101,10 @@ pub trait Resizable {
     fn remove_row(&mut self, row: usize);
     /// Removes a column from a data set by index.
     fn remove_column(&mut self, column: usize);
-    /// Inserts a row to specific by row index.
+    /// Inserts a row at index.
     fn insert_row(&mut self, row: usize);
+    // Inserts column at index.
+    fn insert_column(&mut self, column: usize);
 }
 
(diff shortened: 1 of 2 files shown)


zhiburt:
Hi @IsaacCloos 

> P.S. I'm getting a yellow squiggly that says
> 
> > missing documentation for an associated function

See it has 2 slashes but not 3.

https://github.com/zhiburt/tabled/blob/24f6358508f38885d1a2f441a4c9183221606ca7/tabled/src/records/mod.rs#L106

PR #299, opened 2023-03-07
Split setting

IsaacCloos:
Split checklist:
- [x] code
- [x] tests (please add more as you see fit 😄)
- [x] comments
- [X] example `examples/split.rs`
- [x] readme
- [x] examples/readme

Please give this a spin and let me know what you think. I'm always happy to make any necessary changes to get this right 😄 

---

Also, #290 describes extending this feature to the builder as well. This PR is only related to the setting implementation. If you like how it works I can look at updating the builder next, or it can be a ticket for other contributors to pick up. Who knows!


tabled/src/settings/split/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    pub fn row(index: usize) -> Self {
+        Split {
+            direction: Direction::Row,
+            behavior: Behavior::Zip,
+            index,
+        }
+    }
+
+    pub fn set_behavior(self, behavior: Behavior) -> Self {
```
zhiburt, line 43: Just a thought.

What do you think about not exposing the `Behavior` and create a method `append`/`zip`.

```
    pub fn append(self) -> Self {
        Self { behavior: Behavior::Append, ..self }
    }

   pub fn zip(self) -> Self {
        Self { behavior: Behavior::Zip,  ..self }
    }
```

`zip` method kind of sounds good but `append` might be misleading though....

tabled/src/settings/split/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+                                    (reference_row, reference_column),
+                                );
+                            }
+                        }
+                    }
+                }
+
+                for column in (*index..before_columns).rev() {
+                    records.remove_column(column)
+                }
```
zhiburt, line 86: Maybe move the code in the branches to the functions?
And leave only match in the function.

Likely wont be necessary then?

https://github.com/zhiburt/tabled/blob/0710745eb4c35e1b3018de2fffc62b0c04bd8af0/tabled/src/settings/split/mod.rs#L53-L57

tabled/src/settings/split/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+                                    (reference_row, reference_column),
+                                );
+                            }
+                        }
+                    }
+                }
+
+                for row in (*index..before_rows).rev() {
+                    records.remove_row(row)
+                }
```
zhiburt, line 170: (just a note) You could create a function which would take `impl Iterator<item=usize>` and make the removal (it'd remove duplication in the branch above)

tabled/src/settings/split/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+                                    (reference_row, reference_column),
+                                );
+                            }
+                        }
+                    }
+                }
+
+                for column in (*index..before_columns).rev() {
+                    records.remove_column(column)
+                }
```
zhiburt, line 115: (just a note) You could create a function which would take `impl Iterator<item=usize>` and make the removal (it'd remove duplication in the branch above)

zhiburt (commented):
looks good to me.

zhiburt:
(unrelated)

By the way I guess you've noticed that now all the settings are hidden in a `settings` module so they must be used as `tabled::settings::style::Style`.

Do you find it OK or it's better to provide some often used ones in a `tabled::settings::*`.
Or now I am thinking maybe add a `prelude` for often used settings?

Like `tabled::prelude::*` which would contain things like `Style`,`Padding` etc.

tabled/src/settings/split/mod.rs
IsaacCloos: Great suggestion 😄 

 I'm not sure what to rename append to though. Perhaps `join` or `concatenate`? I want it to be predictable.

IsaacCloos:
> (unrelated)
> 
> By the way I guess you've noticed that now all the settings are hidden in a `settings` module so they must be used as `tabled::settings::style::Style`.
> 
> Do you find it OK or it's better to provide some often used ones in a `tabled::settings::*`. Or maybe add a `prelude` for often used settings?
> 
> Like `tabled::prelude::*` which would contain things like `Style`,`Padding` etc. Though it would bring a bit of duplication...
> 
> I am asking cause, sometimes it's just feels a bit long import to me.

I do think the import tree is kinda long. With the [rust-analyzer](https://rust-analyzer.github.io/) it's not so bad to write it and have the imports generate, but that's only because I am already familiar with the library.

I really like your idea of having a `tabled::prelude::*` section where you can cherry pick the highest traffic settings and features 👍🏻 That could be something good to write about at the top of the readme / quick start guide.

> Though it would bring a bit of duplication...

Smart to be concerned about this, but I can see a quality-of-life improvement here for users!

tabled/src/settings/split/mod.rs
IsaacCloos: https://github.com/zhiburt/tabled/pull/299/files#diff-881689aca954c8702c2bde0c3bf79fb1240972e785c9e672358dcc1407c05e33R204-R212

tabled/src/settings/split/mod.rs
IsaacCloos: https://github.com/zhiburt/tabled/pull/299/files#diff-881689aca954c8702c2bde0c3bf79fb1240972e785c9e672358dcc1407c05e33R194-R202

tabled/src/settings/split/mod.rs
IsaacCloos: https://github.com/zhiburt/tabled/pull/299/files#diff-881689aca954c8702c2bde0c3bf79fb1240972e785c9e672358dcc1407c05e33R80-R192

I'm on the fence about this change. I think I've made a lot more code 😅

tabled/src/settings/split/mod.rs
IsaacCloos: https://github.com/zhiburt/tabled/pull/299/files#diff-881689aca954c8702c2bde0c3bf79fb1240972e785c9e672358dcc1407c05e33R43-R55

tabled/src/settings/split/mod.rs
zhiburt: I guess it's better though (yes.... more lines but seems more clear).
Made a few refactorings 39b628b

zhiburt:
Just made a few experements accoding your comments that it's became longer.

But I've noticed that Split::row doesn't work.

image

1. In `concat` (old `append`) it has trailing columns.
2. `zip` doesn't work at all?

tabled/src/settings/split/mod.rs
zhiburt: concatenate is good?
(shorten to concat?)

IsaacCloos: I like it!

IsaacCloos:
I really like the changes you've made! Especially `ceil_div`, much better now 😄 

---

Now, let's talk about behaviors, predictability, and a new proposal to improve this setting!

> But I've noticed that Split::row doesn't work.

From my testing with the latest changes on this branch, it would seem that `Split::row` is working as I originally intended for both `concat` and `zip`. I will elaborate (also see column reference [here](https://github.com/zhiburt/tabled/issues/290#issuecomment-1450706168) and consider this behavior in the other direction):

```
+---+---+---+---+
| a | b | c | d |
+---+---+---+---+
| e | f | g | h |
+---+---+---+---+
| i | j | k | l |
+---+---+---+---+
| m | n | o | p |
+---+---+---+---+
| q | r | s | t |
+---+---+---+---+
| u | v | w | x |
+---+---+---+---+
| y | z |   |   |
+---+---+---+---+
```
`Split::row(2).zip()                                          Split::row(2).concat() `
example picture

> zip doesn't work at all?

Zip moves by section over direction. Meaning that the provided index determines there should be a cluster made out of every n  number of rows. Those clusters are pushed up behind, one after another, the first cluster. Since there are 7 rows, using `Split::row(2).zip()` causes a `ciel_div` return of 4. In the first column that gives orphan 'y' a new buddy T::default(), and those empty cells in the bottom row act as anchors for new empty cells too.

image

becomes

image

> In concat (old append) it has trailing columns.

The extra columns you are seeing are the two empty cells at the bottom right of the original table anchoring two added T::default() value cells. These cells are added for the same reason `'a'..='z'` split on column 4 created those two empty cells in the first place; the index argument determines the size of the sections your table is divided into. T::default() cells are added as needed to satisfy the new shape.

With all that said, when we settle on an implementation I will create many tests and look over the accuracy of these functions in greater detail.


## Proposal

I think it could improve this feature to have a second behavior for reducing whitespace. By default a split table can, as a final step, check if it has any entirely empty columns/rows and delete those as well as the `cleaned` range. This would resolve this issues you raised in the examples above. You could also opt to keep all rows if desired with a new set of functions like this:

```
// this would be the default behavior
pub fn purge(self) -> Self {
    Self { purge: true, ..self }
}

pub fn retain(self) -> Self {
    Self { purge: false,  ..self }
}
```

zhiburt:
> Zip moves by section over direction. 

Correct.

It works like this in `column` split ( "Zip moves by section over direction. ")

```
 +---+---+                                                                                                 
 | a | b |                                                                                                 
 +---+---+                                                                                                 
 | c | d |                                                                                                 
 +---+---+                                                                                                 
 | e | f |                                                                                                 
 +---+---+                                                                                                 
 | g | h |                                                                                                 
 +---+---+                                                                                                 
 | i | j |                                                                                                 
 +---+---+                                                                                                 
 | k | l |                                                                                                 
 +---+---+                                                                                                 
 | m | n |                                                                                                 
 +---+---+                                                                                                 
 | o | p |                                                                                                 
 +---+---+                                                                                                 
 | q | r |                                                                                                 
 +---+---+                                                                                                 
 | s | t |                                                                                                 
 +---+---+                                                                                                 
 | u | v |                                                                                                 
 +---+---+                                                                                                 
 | w | x |                                                                                                 
 +---+---+                                                                                                 
 | y | z |                                                                                                 
 +---+---+                                                                                                 
 |   |   |                                                                                                 
 +---+---+
```

But I believe it's not what we are getting in row split?
We get.

```
 +---+---+---+---+---+---+---+---+---+---+---+--+---+---+---+--+                                           
 | a | i | q | y | b | j | r | z | c | k | s |  | d | l | t |  |                                           
 +---+---+---+---+---+---+---+---+---+---+---+--+---+---+---+--+                                           
 | e | m | u |   | f | n | v |   | g | o | w |  | h | p | x |  |                                           
 +---+---+---+---+---+---+---+---+---+---+---+--+---+---+---+--+
```

But why not this?

```
 +---+---+---+---+---+---+---+---+---+---+---+---+---+---+
 | a | b | c | d | e | f | g | h | i | j | k | l | m | n |                    
 +---+---+---+---+---+---+---+---+---+---+---+---+---+---+                                           
 | o | p | q | r | s | t | u | v | w | x | y | z |   |   |                                           
 +---+---+---+---+---+---+---+---+---+---+---+---+---+---+
```

> The extra columns you are seeing are the two empty cells at the bottom right of the original table anchoring two added T::default() value cells.

Right

> These cells are added for the same reason 'a'..='z' split on column 4 created those two empty cells in the first place; the index argument determines the size of the sections your table is divided into. T::default() cells are added as needed to satisfy the new shape.

So the reason is right but it's not "the same" as you're saying.
In this case we don't need "satisfy" the shape cause the shape is OK already.
I believe there shall be not 2 trailing columns.

> ## Proposal

If I am correct it wont be necessary,

Though what could be additionally added is logic by which axis we are moving. 
Current logic is based on horizontal axis (from left to right).
We could have an vertical axis option to move (from top to bottom).

But it might be too much.

IsaacCloos:
I like the axis idea!
 ---
 
The reason I think 
> 
```
 +---+---+---+---+---+---+---+---+---+---+---+---+---+---+
 | a | b | c | d | e | f | g | h | i | j | k | l | m | n |                    
 +---+---+---+---+---+---+---+---+---+---+---+---+---+---+                                           
 | o | p | q | r | s | t | u | v | w | x | y | z |   |   |                                           
 +---+---+---+---+---+---+---+---+---+---+---+---+---+---+
```

would be a surprise to the user for a default behavior

IsaacCloos:
I think it's easier to visualize the similarity with how they work by changing the `table_1` split column index to 7. You can see some funny (and I think necessary) business happening in both directions for tables_[2,3,4,5]

IsaacCloos:
Thought exercise about how and when to trim empty rows:
```
let table = Table::from_iter([vec![String::new(); 26]]);
```

should splitting this in any direction with any behavior result in an empty (0 by 0) table? Who's to say whether the cells being blank is indicative of uselessness to the user? I'm thinking they should be able to opt to retain those empties, and even if they want them removed the process should only consider hanging elements such as blank ceil_div + 1 cells.

This rule could be stated as *"Blank cells don't have the ability to anchor added rows or columns to satisfy a new shape"*

- This rule would lead to "intra blanks" being kept while bouncing newly created blanks and their anchors. 
- This rule would directly address the blank columns you've identified in your most recent picture.

Thoughts?

IsaacCloos:
```
fn split_row_zip<R>(records: &mut R, columns: usize, rows: usize, index: usize)
where
    R: Resizable + ExactRecords,
{
    let per_column = ceil_div(rows, index);

    let mut thrown_out_sections = 0;

    for col in 0..columns {
        let ref_col = col * per_column - thrown_out_sections;

        for section in 1..per_column {

            let target_col = ref_col + section;
            records.insert_column(target_col);

            for target_row in 0..index {
                let ref_row = target_row + (section * index);

                if ref_row < rows {
                    if section == per_column - 1
                        && records.get_cell((ref_row, ref_col)).as_ref() == ""
                    {
                        records.remove_column(target_col);
                        thrown_out_sections += 1;
                        println!("blank count: {thrown_out_sections}");
                        continue;
                    }
                    records.swap((target_row, target_col), (ref_row, ref_col));
                }
            }
        }
    }

    clean_rows(records, (index..rows).rev());
}
```

Threw together this updated function during my lunch hour. This would specifically bounce final, blank cells, and prevent them from creating new columns in the output. It makes this:
```
+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
| a | i | q | y | b | j | r | z | c | k | s | d | l | t |
+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
| e | m | u |   | f | n | v |   | g | o | w | h | p | x |
+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
```

IsaacCloos:
Phew! Who would have expected this feature to spawn such great conversations 😄 

These latest changes address points you've made about blank/trailing columns/rows:
- empty rows and cells are bounced (please run the latest example to see this in action)
- the function is now attempting to fit non-blank cells into as few sections as possible with a minimum of 1 (the current example also demonstrates this behavior with an empty table). I'm curious what you'll think of this.

Of course still subject to change with new feedback 👍🏻

zhiburt:
As I see it works (at least in example :))
But do you think it shall be this way that table with empty cells has a different results from not empty table? (I mean your 1st and 2nd table, as I understand you're explicitly put it there)

image

I want to believe it must act in the same way?

IsaacCloos:
These latest changes are my attempt to greatly reduce the duplicate code! 

Futhermore, I figured, forget everything I said before about complicated rules for eliding anchored sections and whatnot. 

My new, simple, and hopefully predictable logic is:
> if a newly created section would be blank, remove it, and give the user the option to NOT do that if they want

I hope this is flexible, while being concise by default with no trailing columns/rows.

Thanks again for working with me on this 😄

zhiburt:
Run your old example;

```rust
    let table = Table::from_iter([vec![String::new(); 26]]);
    let table1 = table.clone().with(Split::column(12)).clone();
    let mut table = col[table, table1].with(Style::blank()).to_string();

    println!("{table}");
```

```
'attempt to subtract with overflow', /home/maxim/projects/tabled/tabled/src/settings/split/mod.rs:122:25
```

___________

I think that `retain` must be a default.

But still there's this issue with additional columns.
We have already disscussed it.

image
image

I'd expect to see no more than we have.

image

(I'd not be very worried with performance here) So the easiest fix would be call `remove_column` a necessary amount.
As sometimes we would need to push additional columns (as you do)

IsaacCloos:
thank you for finding this empty table bug! I will push a fix

tabled/src/settings/split/mod.rs
```diff
@@ -0,0 +1,333 @@
...
+                        inner_index,
+                        inner_index + outer_index * secondary_length - filtered_sections,
+                    ),
+                    Behavior::Zip => (
+                        inner_index,
+                        usize::checked_sub(outer_index * sections_per_direction, filtered_sections)
+                            .unwrap_or(0),
+                        outer_index * sections_per_direction + inner_index - filtered_sections,
+                    ),
+                };
```
zhiburt, line 274: maybe make it a function?

tabled/src/settings/split/mod.rs
```diff
@@ -0,0 +1,333 @@
...
+                            format_position(direction, from_primary_index, from_secondary_index);
+                        if records.get_cell(from_position).as_ref() != "" {
+                            section_is_empty = false;
+                        }
+                        records.swap(
+                            from_position,
+                            format_position(direction, to_primary_index, to_secondary_index),
+                        );
+                    }
+                }
```
zhiburt, line 300: maybe make it a function?

zhiburt (commented):
So as long as you're good; I am good; just let me know;

__________________

I've left a small comment about maybe move some code into function;
It's often might be a code smell where you use a comment whereas you could use a function.
I just wanted to highlight it but if you consider it's better be left as it is that's good.

Also maybe add a few tests for empty table? (`Builder::new().build()`)

tabled/Cargo.toml
```diff
@@ -239,6 +239,11 @@ name = "alphabet"
 path = "examples/alphabet.rs"
 required-features = ["std"]
 
+[[example]]
+name = "split"
+path = "examples/split.rs"
+required-features = []
```
zhiburt, line 245: Add here `"macros"` to fix a CI.

IsaacCloos: Thank you!
https://github.com/zhiburt/tabled/blob/f5c56a4b50a5e34a04035bc19cefb7e088b22a8f/tabled/Cargo.toml#L242-L243

I was thinking that this block of code meant that the macros feature was enabled by default, and was scratching my head why those checks were still failing 😅

IsaacCloos:
> Also maybe add a few tests for empty table? (Builder::new().build())

There are tests for empty tables, usize 0 arguments, and splitting on indexes greater than the size of the table. However, I do think I need to add some more; cover more ground to make sure this thing is ready 👍🏻

tabled/src/settings/split/mod.rs
IsaacCloos: I know you already flagged this as resolved, but I followed your advice and I think it's cleaner now 😄

IsaacCloos:
What a journey! 

Please give this a hard look and don't hold back on the criticism 😄 I'd be happy to make any necessary changes ✏️

zhiburt:
I'll take a closer look a bit later.

> I'd be happy to make any necessary changes pencil2

Just got this this in the mind.
Remember this comment https://github.com/zhiburt/tabled/issues/290#issuecomment-1447049886?

I am thinking wouldn't it be cool to be able to do this.

```rust
let table = ['a'..='z'].iter()
    .collect::<UniformBuilder>()
    .columns(12)
    .build()
```

It would build a table with 12 columns from an iterator.

To do that it'd be necessary to add a new builder to `/tabled/tabled/src/builder/`,
and implement `FromIterator` for it. (also columns + build) method.

I am not sure it might be unrelated a bit.
And cause the PR already got a bit too long (at least the discussion it might be better be seppareted).

But let me know what do you think.

zhiburt:
Looks good to me;
Thanks for the contribution.

Issue #26, opened 2021-10-14
Hyperlinks break width detection


maxbla:
Following instructions in [this gist](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda), the way to print a hyperlink in the terminal with rust is something like `format!("\x1b]8;;{}\x1b\\{}\x1b]8;;\x1b\\", display, url)` which displays in the terminal as having width `display.len()`, but tabled appears to think its display width is approximately `14+display.len()+url.len()`. This and others are listed in [the operating system command section](https://en.wikipedia.org/wiki/ANSI_escape_code#OSC_(Operating_System_Command)_sequences) of the relevant wikipedia page.

An example of the results
```
┌─────────────────────────────────────────────────────────────────────────────┬────────────────────┐
│                                    name                                     │     similarity     │
├─────────────────────────────────────────────────────────────────────────────┼────────────────────┤
│ Person1 │        100         │
│         Person2         │ 65.38461538461539  │
│      Person3      │ 62.962962962962955 │
│    Person4    │ 62.962962962962955 │
│       Person5       │ 60.714285714285715 │
└─────────────────────────────────────────────────────────────────────────────┴────────────────────┘
```

after glancing at some docs, I think the underlying issue is with ansi-parser

maxbla:
I don't like just `ansi` as a feature name because that's a whole standards body, who happen to have a relevant standard here. It isn't immediately clear to a library user which standard `ansi` refers to (ANSI C is another popular one). I think `color` is an okay name, since it basically communicates what is happening. It doesn't say exactly, but if you know how terminals work, you can guess that's what you need. `terminal-escapes` would be more accurate, but is longer and a breaking change.

As for whether the feature should be enabled by default, I think not. Of your 5 listed reverse-dependencies on crates.io, only one uses the color feature (qcow-cli) so you would be adding a significant overhead for no benefit to most of your users. You could re-evaluate your reverse dependencies later, as they might change in the future.

PR #307, opened 2023-03-14
Better discoverability for `color` feature in README.md

jondot:
Added another example, a link and hopefully larger surface area for people searching for a solution for working with colors inside a table. Took me quite a while to find that feature as i've searched for:

* broken ansi
* color in cell
* bad layout

and other search terms, just to hit issue #26 which deals with _hyperlinks_ ansi, but describes the problem well.

After I've used the color feature, everything works like magic!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1358,8 +1358,24 @@ println!("{}", table);
 
 ### Color
 
-The library doesn't bind you in usage of any color library but to be able to work correctly with color input you should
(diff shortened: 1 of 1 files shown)


zhiburt:
Hi @jondot 

Yes.....the documentation is something we shall iterate on.

Thanks for the contribution;
Let me know if you have any other suggestions.

Take care.

Issue #296, opened 2023-03-01
Example output migration


IsaacCloos:
>I wonder what do you think about putting the results of examples right to the head comment of the example?
We have  the results in examples/README.md but I wonder if it would be any useful to have it right there.

_Originally posted by @zhiburt in https://github.com/zhiburt/tabled/issues/294#issuecomment-1450169714_

zhiburt:
I am actually thinking maybe adding a description to the example would be more valuable e.g.

> This example demonstrates padding changes of a particular cell.
> `Padding` is stand for left/right/top/bottom inner indent of a cell.

IsaacCloos:
I'd be happy to make these changes 😄  

Do you approve of these new goals for this issue:
- [ ] remove [/tabled/examples/readme](https://github.com/zhiburt/tabled/blob/master/tabled/examples/README.md)
- [ ] update all examples with brief descriptions (like you demonstrated in this [comment](https://github.com/zhiburt/tabled/issues/296#issuecomment-1458189149))

zhiburt:
> remove [/tabled/examples/readme](https://github.com/zhiburt/tabled/blob/master/tabled/examples/README.md)

Welllllll I guess we better do not delete it:)
I believe it's quite informative; to see what kind of tables we may build.

I guess my original comment was **misleading**;
I guess the example files better be short (some of them are quite big which might be hard to get already).
And why having output in comment if we could do an assert in the code so it could be used as test as well?

__________________________

1. the idea to have a descriptions of; "What the example does and what it tries to accomplish" is good.
2. the output example in a top comment might be unnecessary? @IsaacCloos 
3. do not delete the readme :)

IsaacCloos:
I will write up a short form explanation of every example in the project ✏️ 

Thanks for clarifying 😄

PR #314, opened 2023-03-21
Example description revisions

IsaacCloos:
# Addressing #296

I've updated a handful of the examples, and would love some feedback before I proceed with the rest. I'm _attempting_ to keep the descriptions brief while identifying all the major ideas of each example and providing links where appropriate to further contextualize libraries referenced.

# Progress

- [X] csv_to_table
- [X] json_to_table
- [x] papergrid
- [x] static_table
- [x] table_to_html
- [x] tabled
    - [X] derive
- [X] tabled_derive

# Omitted [^1]

 - use_in_docs `static_table`
 - show `tabled`
 - terminal_size `tabled`

# Criticism Welcome 😄 
If you have any preferences whatsoever, please let me know and I will change what we've got so far. Or if you'd rather leave the examples as-is that's good too 👍🏻 

[^1]: tests not `cargo run --example` accessible

### DIFF
--- a/csv_to_table/examples/csv_table.rs
+++ b/csv_to_table/examples/csv_table.rs
@@ -3,7 +3,7 @@
 //! `cargo run --example csv_table`
 //!
 //! This example demonstrates reading a csv string to a [`Table`] struct.
-//! 
+//!
 //! ---
 //!
 //! * Note the necessary step of representing the string as a byte array.
--- a/csv_to_table/examples/csv_table.rs
+++ b/csv_to_table/examples/csv_table.rs
@@ -1,5 +1,12 @@
-//! The example can be run by this command
+//! This example can be run with the following command:
+//!
 //! `cargo run --example csv_table`
+//!
+//! This example demonstrates reading a csv string to a [`Table`] struct.
+//! 
+//! ---
+//!
+//! * Note the necessary step of representing the string as a byte array.
 
 fn main() {
     let syscalls = "\
--- a/csv_to_table/examples/iter.rs
+++ b/csv_to_table/examples/iter.rs
@@ -2,13 +2,13 @@
 //!
 //! `cargo run --example iter`
 //!
-//! This example demonstrates an alternative parsing option for csv-to-table translations.
+//! This example demonstrates an alternative parsing option for [`csv_to_table`] translations.
 //!
 //! ---
 //! 
-//! * The formatting feature `sniff` is used to demonstrate how many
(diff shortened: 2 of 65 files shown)


zhiburt:
It feels like in some cases your description is more expressive then we have, as docs.

> ```rust
> //! * The [`IterTable::sniff()`] formatting function is used to control how many
> //! rows will be considered in determining column widths. Since the default sniff value is 1000,
> //! this is helpful for controlling large dataset display outputs.
> ```

vs 

```rust
/// Limit an amount of rows will be read for dimension estimations.
```

I do think we can go with what you'll chose the right one.

IsaacCloos:
Ok, Thanks for the feedback! I'll work through the rest of these shortly 😄

IsaacCloos:
I hope the goal of making the example browsing experience more educational for users was achieved.

# Criticism Welcome ❗ 
Please let me know if you'd like to see any changes 😄 
... or even rejecting the work I understand too. This is quite a lot to push through in one go, I may have gotten carried away writing these up ha!

zhiburt:
Seems good to me.
As I said previously some comments have more information than the code docs.

I'll merge, thanks for such a heavy lifting.

___________________________________

What do you think about this section.
I was thinking previously it might be useful for someone who is not used to rust yet so they know how to run it.
But maybe it's just redundant?

```
//! This example can be run with the following command:
//!
//! `cargo run --example static_table`
```

IsaacCloos:
> What do you think about this section. I was thinking previously it might be useful for someone who is not used to rust yet so they know how to run it. But maybe it's just redundant?
> 
> ```
> //! This example can be run with the following command:
> //!
> //! `cargo run --example static_table`
> ```

I agree. I've checked around a few other Rust projects, and haven't found another that specifies the command. 

However, there are a few examples that ask the user to write custom inputs outside of cargo, so perhaps keeping those would be helpful. e.g.
https://github.com/zhiburt/tabled/blob/5b254f3808bdc14cf97ffaea301b3e2e845b207a/tabled/examples/shadow.rs#L1-L3

or honestly these examples could be simplified? I don't see how using the Stdin adds to Shadow setting example, but I didn't think these doc revisions was the right place to challenge it.

This one I think is worth keeping for sure, since using Stdin is the whole point of the example.
https://github.com/zhiburt/tabled/blob/5b254f3808bdc14cf97ffaea301b3e2e845b207a/json_to_table/examples/reader.rs#L1-L3

PR #315, opened 2023-03-22
fix(macro hygiene): use the absolute path in Tabled derive macro

zjp-CN:
and the caller doesn't need to import Tabled trait


zhiburt:
Hi @zjp-CN thanks for the commitment.

Indeed you're correct.

PS: I actually encountered this issue a few days ago, and was perplexed... and didn't payed a necessary attention.

PR #364, opened 2023-10-26
tabled/ Fix IndexBuilder on empty tables (#363)

### LINKED ISSUE #363: `table.is_empty()` on tables with only a header row
Hi, I run into a situation where I need to check if a table with a header row is empty. I notice that `table.is_empty()` returns false if the table has no data rows but merely a header row. However the function returns true if the table is indexed:

```rust
use tabled::{builder::Builder, settings::Style};

fn main() -> Result<()> {
    let mut builder = Builder::default();
    builder.set_header(["color", "vegetable", "contains"]);
    let builder = builder.index().column(1).name(None);
    let table = builder.build();
    assert!(table.is_empty());

    let mut builder = Builder::default();
    builder.set_header(["color", "vegetable", "contains"]);
    let table = builder.build();
    assert!(!table.is_empty());

    Ok(())
}
```
The difference seems to be caused by that `IndexBuilder` takes ["data length"](https://github.com/zhiburt/tabled/blob/master/tabled/src/builder/index_builder.rs#L229-L230) into consideration. Is this difference by-design or should non-indexed tables do the same too?

pjhades:
There's a minor behaviour difference between how `Builder` and `IndexBuilder` handles tables that have no data rows but a header row. In that case, `Builder` will regard the header row itself as data, but `IndexBuilder` will not. This difference shows up when `table.is_empty()` is called on the resulting tables: such tables will have different shapes.

This patch fixes the process when an `IndexBuilder` is converted back to a `Builder`. With this patch, a default `Builder` will be returned only if the `IndexBuilder` has neither data rows nor a header row.

Unit tests are added accordingly.

Fix #363 

Now this code works:
```rust
use tabled::builder::Builder;

fn main() -> Result<()> {
    let mut builder = Builder::default();
    builder.set_header(["color", "vegetable", "contains"]);
    let builder = builder.index().name(None);
    let table = builder.build();
    assert!(!table.is_empty());
    assert_eq!(table.shape(), (1, 4));

    let mut builder = Builder::default();
    builder.set_header(["color", "vegetable", "contains"]);
    let table = builder.build();
    assert!(!table.is_empty());
    assert_eq!(table.shape(), (1, 3));

    Ok(())
}
```

### DIFF
--- a/tabled/src/builder/index_builder.rs
+++ b/tabled/src/builder/index_builder.rs
@@ -246,7 +246,8 @@ impl From<IndexBuilder> for Builder {
 }
 
 fn build_index(mut b: IndexBuilder) -> Builder {
-    if b.index.is_empty() {
+    // we can skip the conversion if this builder has neither data rows nor header row
+    if b.index.is_empty() && matrix_count_columns(&b.data) == 0 {
         return Builder::default();
     }
 
--- a/tabled/tests/core/index_test.rs
+++ b/tabled/tests/core/index_test.rs
@@ -157,7 +157,7 @@ test_table!(
 );
 
 test_table!(
-    builder_index_invalid_dosnt_panic,
+    builder_index_invalid_doesnt_panic,
     Builder::default().index().column(100).build(),
     ""
 );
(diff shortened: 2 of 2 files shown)


zhiburt:
Thanks for your contribution @pjhades

Feel free to open an issue if you see something wrong/odd to your opinion.

It hopefully will be released after some refactoring I need to finish.

Take care.

PS: the `clippy` warning is about another thing

Issue #372, opened 2023-11-13
Refactor `Builder` methods


zhiburt:
1. Add `remove_row` `remove_column` methods.
2. Rename `clean` method
3. Replace `clean` with a real `clean` which would remove all the data
4. (probably) Rename `push_record` to `push_row`
5. Add `push_column`
6. Add `insert_column`
7. Additionally it would be nice to have a consistent interface with `IndexBuilder` (One uses `&mut self` but another `self`)

CouldBeFree:
Regarding the first point. It looks like the `remove_row` and `remove_column` methods should accept the index.

CouldBeFree:
Yes, you can assign it to me. I am new to Rust, so If you don't mind I will ask stupid questions during the task.

CouldBeFree:
@zhiburt how can I test the methods which I added via impl Table?

zhiburt:
> @zhiburt how can I test the methods which I added via impl Table?

There's many ways.
The smartest is probably to write actual tests, (including the fact that's relatively straightforward).

See, you could add one more

https://github.com/zhiburt/tabled/blob/74fcf7ade12d3ba5f461f67536f1a32245ef2192/tabled/tests/core/builder_test.rs#L9-L25

CouldBeFree:
What do you mean by the third point?

> Replace clean with a real clean which would remove all the data

As I see this method clear columns and rows. What else is this function supposed to do? 
```
pub fn clean(&mut self) -> &mut Self {
   self.clean_columns();
   self.clean_rows();
   self
}
```

zhiburt:
> As I see this method clear columns and rows. What else is this function supposed to do?

I did a mistake actually....I was thinking it's called clean.... https://doc.rust-lang.org/std/vec/struct.Vec.html#method.clear

So I think it can stay untouched.
We just need to add one more method `clear` which would remove all the data, but left headers.

zhiburt:
> I have a question regarding remove_row and remove_column, as far as those functions update some vec by the given index, should they panic if the index is more than vec length or just do nothing in this case?

Good question;
Yes.

We just need to add a doc comment about it.

CouldBeFree:
> > I have a question regarding remove_row and remove_column, as far as those functions update some vec by the given index, should they panic if the index is more than vec length or just do nothing in this case?
> 
> Good question; Yes.
> 
> We just need to add a doc comment about it.

I suppose that you mean this -> `/// # Panics`

[...]

PR #375, opened 2023-11-22
Refactor builder methods

CouldBeFree:


zhiburt:
- [x] Add remove_row remove_column methods.
- [x] Rename clean method
- [x] Add push_column
- [x] Add insert_column

ref #372

PR #392, opened 2024-02-23
tabled/ fix Builder::push_column cannot create new rows

### LINKED ISSUE #391: Builder::push_column cannot create new rows
https://github.com/zhiburt/tabled/blob/4ab897e7e4e6449dd51d3297229f072702dfc49d/tabled/src/builder/table_builder.rs#L281-L290

In the above code, `row` is not inserted into the table

zxch3n:
- Fixed  #391

### DIFF
--- a/tabled/src/builder/table_builder.rs
+++ b/tabled/src/builder/table_builder.rs
@@ -287,6 +287,7 @@ impl Builder {
             }
 
             row.push(text);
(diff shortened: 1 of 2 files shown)


zhiburt:
Thanks,

You were right about my take on the empty iterator.

Take care

PR #397, opened 2024-03-24
tabled_derive/ Fix issue with Cow usage

### LINKED ISSUE #395: "the trait `From<std::path::Display<'_>>` is not implemented for `Cow<'_, _>`"
The following example results in an error spelled out in the title:

```rust
#[derive(tabled::Tabled)]
struct MyStruct<'a> {
    #[tabled(display_with("std::path::Path::display"))]
    path: &'a std::path::Path,
}
```

`tabled` only says that it requires types to be `Display`, but here is it also trying to construct a `Cow` of something from that `Display` impl.

zhiburt:
Seems to work @elenakrittik?

https://github.com/zhiburt/tabled/blob/7ec7973c1814be4cd909b399357c36d45464c1a7/tabled/tests/derive/derive_test.rs#L1084-L1091

close #395

### DIFF
--- a/json_to_table/src/table/collapsed_table.rs
+++ b/json_to_table/src/table/collapsed_table.rs
@@ -108,7 +108,7 @@ fn _collapsed_table(val: &Value, cfg: &Config, dims: &Dimensions, ctx: PrintCont
 }
 
 fn generate_vertical_array(
-    list: &Vec<Value>,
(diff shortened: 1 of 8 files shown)


elenakrittik:
Yep, works for me with this patch. Thanks

Issue #396, opened 2024-03-22
Add a new `Tabled` attribute `format = ""` as a shortcut for a `display_with`


zhiburt:
I guess it's make sense to do.
To reduce a boilerplate for common simple cases.

```rust
#[derive(Tabled)]
struct Motorcycle {
  weight: usize,
  #[tabled(format = "{} cc")]
  cc: usize,
}
```

I think a 2nd version could exist, to being able to use a self as argument.

```rust
#[derive(Tabled)]
struct Motorcycle {
  weight: usize,
  #[tabled(format("{}/{} cc/kg", self.cc, self.weight))]
  cc: usize,
}
```

thomassimmer:
Hello,

I started working on this. Just need to make tests work for enums. Could you assign this to me ?

Cheers,
Thomas

zhiburt:
@thomassimmer 

Thanks

If you have any question please ping me.

But overall it shall be very very very similar to `display_with` implementation.
Don't be frightened by quite bad quality code at `tabled_derive/src/lib.rs`.

Take care.

zhiburt:
closed by #398

PR #398, opened 2024-03-25
Tabled attribute format

thomassimmer:
Hello,

This is for #396 .

I hope it's fine, it was probably a bit too difficult for me as a first issue. I tried to do something very similar to display_with but I encounter some difficulties with properties and nested structs/enums. I would be happy to see what corrections you can make to improve it.

By the way, thanks a lot for the library! :)

FYI, I needed that for this project : https://github.com/thomassimmer/NeoPass

Cheers

### DIFF
--- a/README.md
+++ b/README.md
@@ -1592,6 +1592,36 @@ impl MyRecord {
 }
 ```
 
+To reduce boilerplate code, one can also achieve this using the `format` attribute within `#[derive(Tabled)]`.
+
+```rust
+use tabled::Tabled;
+
+#[derive(Tabled)]
+pub struct Motorcycle {
+    weight: usize,
+    #[tabled(format = "{} cc")]
+    cc: usize,
+}
+```
+
+In the above example, the cc field will be formatted using the specified format string "{} cc", where {} is replaced with the value of cc.
+
+Just like with `display_with` attribute, you can pass arguments for more complex formatting scenarios:
+
+```rust
+use tabled::Tabled;
+
+#[derive(Tabled)]
+pub struct Motorcycle {
+    weight: usize,
+    #[tabled(format = "{}/{} cc/kg", "self.cc", "self.weight")]
+    cc: usize,
+}
+```
+
+In this case, the cc field will be formatted using the format string "{}/{} cc/kg", and {} will be replaced with the values of cc and weight, respectively.
+
 ### Inline
 
 It's possible to inline internal data if it implements the `Tabled` trait using `#[tabled(inline)]`.
--- a/README.md
+++ b/README.md
@@ -1615,7 +1615,7 @@ use tabled::Tabled;
 #[derive(Tabled)]
 pub struct Motorcycle {
     weight: usize,
-    #[tabled(format = "{}/{} cc/kg", "self.cc", "self.weight")]
+    #[tabled(format = "{}/{} cc/kg", self.cc, self.weight)]
     cc: usize,
 }
 ```
--- a/tabled/tests/derive/derive_test.rs
+++ b/tabled/tests/derive/derive_test.rs
@@ -227,22 +227,22 @@ mod tuple {
     );
 
     test_tuple!(
-        format,
-        t: { u8 #[tabled(format = "foo")] sstr },
-        init: { 0 "v2" },
-        expected: ["0", "1"], ["0", "foo"],
-    );
-
-    test_tuple!(
         format_1,
         t: { u8 #[tabled(format = "foo {}")] sstr },
         init: { 0 "v2" },
         expected: ["0", "1"], ["0", "foo v2"],
     );
 
-    // todo : self is the tuple here. It should be the sstr element instead.
+    test_tuple!(
+        format_2,
+        t: { u8 #[tabled(format = "foo {:?}")] sstr },
+        init: { 0 "v2" },
+        expected: ["0", "1"], ["0", "foo \"v2\""],
+    );
+
+    // todo : self represents the tuple here. It should be the sstr element instead.
     // test_tuple!(
-    //     format_2,
+    //     format_3,
     //     t: { u8 #[tabled(format("foo {} {}", 2, self))] sstr },
     //     init: { 0 "v2" },
     //     expected: ["0", "1"], ["0", "foo 2 v2"],
--- a/tabled/tests/derive/derive_test.rs
+++ b/tabled/tests/derive/derive_test.rs
@@ -226,6 +226,28 @@ mod tuple {
         }
     );
 
+    test_tuple!(
+        format,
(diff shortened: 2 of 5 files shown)


tabled_derive/src/lib.rs
```diff
@@ -618,9 +683,26 @@ fn merge_attributes(attr: &mut Attributes, global_attr: &StructAttributes) {
...
+            let property_name = syn::Ident::new(val, proc_macro2::Span::call_site());
+
+            // We find the field in the neighbours instead of taking self, which would be the top object.
+            // This is for nested formatting.
+            for (i, field) in fields.iter().enumerate() {
+                let field_name_result = field_name(i, field);
+                if field_name_result.to_string() == val.to_string() {
+                    return quote! { #field_name_result };
+                }
+            }
```
zhiburt, line 703: Why exactly we need it?

(SRE, just cant crack it fastly)

tabled_derive/src/lib.rs
```diff
@@ -477,6 +531,17 @@ fn use_function_no_args(function: &str) -> TokenStream {
     }
 }
 
+fn use_format(args: &TokenStream, custom_format: &str) -> TokenStream {
+    quote! { format!(#custom_format, #args) }
+}
+
+fn use_format_no_args(custom_format: &str, field: &TokenStream) -> TokenStream {
+    if custom_format.contains("{}") {
```
zhiburt, line 539: just a strange if :sweat_smile:

zhiburt: I guess it's error prone,
I'd guess we could just check arguments length and that's it?

Yes it would probably compile the code, I think.
But got `clippy` issue.

tabled_derive/src/lib.rs
```diff
@@ -477,6 +531,17 @@ fn use_function_no_args(function: &str) -> TokenStream {
...
+        return quote! { format!(#custom_format, #field) };
```
zhiburt, line 540: Is this always will pass a targeted field as first argument to `format!` while ignoring other arguments, in case someone "forgot" `{}` the braces?

I guess so.

I very much believe could be triggered by the test I've described.

__________

See the above comment.
I consider it can be removed if that will be taken in mind.

tabled/tests/derive/derive_test.rs
```diff
@@ -226,6 +226,28 @@ mod tuple {
         }
     );
```
zhiburt, line 228: I'd add tests with debug braces `{:?}`.

tabled/tests/derive/derive_test.rs
```diff
@@ -226,6 +226,28 @@ mod tuple {
...
+    );
+
+    test_tuple!(
+        format_1,
+        t: { u8 #[tabled(format = "foo {}")] sstr },
+        init: { 0 "v2" },
+        expected: ["0", "1"], ["0", "foo v2"],
+    );
+
+    // todo : self is the tuple here. It should be the sstr element instead.
```
zhiburt, line 243: It's a good call to test `self` but what's the issue, exactly?

README.md
```diff
@@ -1592,6 +1592,36 @@ impl MyRecord {
 }
 ```
 
+To reduce boilerplate code, one can also achieve this using the `format` attribute within `#[derive(Tabled)]`.
```
zhiburt, line 1595: Thanks for README update.

note: could be also updated in `tabled` documentation itself.

README.md
```diff
@@ -1592,6 +1592,36 @@ impl MyRecord {
...
+
+Just like with `display_with` attribute, you can pass arguments for more complex formatting scenarios:
+
+```rust
+use tabled::Tabled;
+
+#[derive(Tabled)]
+pub struct Motorcycle {
+    weight: usize,
+    #[tabled(format = "{}/{} cc/kg", "self.cc", "self.weight")]
```
zhiburt, line 1618: I guess you shall have this instead? `#[tabled(format = "{}/{} cc/kg", self.cc, self.weight)]`

zhiburt (changes requested):
I am surprised that you got this fight :smile: 

So overall pretty decent. 

I'd have some comments according to style.
But I have them against my own code, and the current state of `tabled_derive` in particular.... :disappointed_relieved: 

Let me know what you think and whether,
you will be able to address my comments.

Take care

zhiburt:
And yes, don't worry about merge changes.

zhiburt:
Am I wrong that with your changes `display_with` will also support arguments like `self.field1`.
I bet I am.

And I bet it's good.

Just need to document and create tests.

tabled_derive/src/lib.rs
thomassimmer: Passing the fields like this, we can associate the right properties. Otherwise, self could represent the highest level struct if we are trying to format the property of a nested struct.

Is it more clear? The test "format" on enums is verifying this issue.

tabled_derive/src/lib.rs
thomassimmer: I totally agree, thank you!

tabled_derive/src/lib.rs
thomassimmer: It's removed.

tabled/tests/derive/derive_test.rs
thomassimmer: done 👍

README.md
thomassimmer: Yes, thank you!

thomassimmer:
> Am I wrong that with your changes `display_with` will also support arguments like `self.field1`. I bet I am not.
> 
> And I bet it's good.
> 
> Just need to document and create tests.

You are right! I added one test, let me know if it's fine.

zhiburt:
Hey @thomassimmer 

I'll merge it,
will add a few more things,
and release hopefully soon.

But could you let me know your thoughts on this `tuple` thing?

thomassimmer:
> Hey @thomassimmer
> 
> I'll merge it, will add a few more things, and release hopefully soon.
> 
> But could you let me know your thoughts on this `tuple` thing? It seems perfectly good to me.
> 
> ```rust
>     #[test]
>     fn format3() {
>         #[derive(Debug, Tabled)]
>         struct StructName(u8, #[tabled(format("foo {} {:?}", 2, self))] String);
> 
>         let value = StructName(0, String::from("string"));
> 
>         assert_eq!(value.fields(), vec!["0", "foo 2 StructName(0, \"string\")"]);
>         assert_eq!(StructName::headers(), vec!["0", "1"]);
>     }
> ```
> 
> Thanks for contribution. And take care.

Hi @zhiburt ,

What you said on tuples sounds good to me too!

Thank you for merging and take care.

Issue #399, opened 2024-03-30
Using Span::column() with the Style::modern() causes virtual artifact on the intersection in the table


fbonin:
Using Span::column() with the Style::modern() causes virtual artifact on the intersection in the table

### To reproduce : 

```
let mut builder = Builder::default();

let mut title = Vec::with_capacity(1);
title.push("Title");
builder.push_record(title);

let mut header = Vec::with_capacity(4);
header.push("header-1");
header.push("header-2");
header.push("header-3");
header.push("header-4");
builder.push_record(header);

let mut record = Vec::with_capacity(4);
record.push("value-1");
record.push("value-2");
record.push("value-3");
record.push("value-4");
builder.push_record(record);

let mut table = builder.build();
table    
.modify((0, 0), Span::column(4))
    .modify((0, 0), Alignment::center())
    .with(Style::modern())
    .to_string();
println!("{}", table);
```

### Artifacts on the header at the intersection : 

<img width="362" alt="image" src="https://github.com/zhiburt/tabled/assets/732394/a1aec7c6-9368-4bd1-8a3a-1fa8d7edc70c">

### Expected : 

<img width="359" alt="image" src="https://github.com/zhiburt/tabled/assets/732394/e91c14c0-988c-4d1b-ad9b-69429313e0c3">

zhiburt:
Hi @fbonin 

That's actually by design.

Try to use `tabled::settings::style::BorderSpanCorrection` afterwards.

Let me know if it helped.

_________________

Yes I've just noted it's not stated in `README.md`.
If you want to you can open a PR

Take care

fbonin:
Thank you !

mkatychev:
Just a heads up: this doesn't work on all styles:
```rust
use tabled::{
    settings::{
        style::{BorderSpanCorrection, Style},
        Modify, Span,
    },
    Table,
};

fn main() {
    let data = vec![("09", "June", "2022"), ("10", "July", "2022")];

    let mut table = Table::new(&data);
    table
        .with(Modify::new((0, 0)).with("date").with(Span::column(3)))
        .with(BorderSpanCorrection);
    println!("{}", table.to_string());
    table.with(Style::modern());
    println!("{}", table.to_string());
}
```

```
$ cargo run

+------------------+
| date             |
+----+------+------+
| 09 | June | 2022 |
+----+------+------+
| 10 | July | 2022 |
+----+------+------+
┌────┬──────┬──────┐
│ date             │
[...]

PR #402, opened 2024-04-20
doc: add documentation for issue #399

fbonin:
Added documentation related to issue #399

### DIFF
--- a/README.md
+++ b/README.md
@@ -1343,7 +1343,7 @@ The resulting table would be the following.
 
 ### Span
 
-It's possible to set a horizontal(column) span and vertical(row) span to a cell.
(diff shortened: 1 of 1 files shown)


zhiburt:
Hi @fbonin 

I do think it's a good proposition to notice it.

But what about making an example of `BorderSpanCorrection` usage?
You could rely on already existing example at https://github.com/zhiburt/tabled?tab=readme-ov-file#horizontal-span
And show the result with `BorderSpanCorrection` used.

What do you think?

zhiburt:
Hey @fbonin 
Sorry that it took that long;

Thanks for your contribution.
Take care.

PR #404, opened 2024-06-02
Fix for cargo clippy

akiomik:
This PR fixes the issue of clippy failing in CI.

As noted in the comments, I've temporarily disabled `assiginig_clones` due to a bug in which clippy does not respect the version of rust. (`assiginig_clones` is actually available since rust 1.63.0. Currently the [`rust-version` in `Cargo.toml` is 1.61.0](https://github.com/zhiburt/tabled/blob/f454b5ab6efd2f2a6b3c3a416098d4fd1347315d/tabled/Cargo.toml#L14).)

ref: https://github.com/rust-lang/rust-clippy/issues/12502


zhiburt:
Hey @fbonin 
Sorry that it took that long;

Thanks for your contribution.
Take care.
Once again

PR #406, opened 2024-06-09
add `Clone`, `Copy` to `tabled::settings::object::Cell`

Alexdelia:
also update `papergrid` `Position` `derive`


zhiburt:
Yes; was missed.

Thanks.

Take care

zhiburt:
> it compiled because tabled is using the crate registry version of papergrid, and not the local one

Yes :smile: 
To be honest I was thinking about it; and decided not to check; was thinking maybe something was changed in a latest rust versions :)

PR #408, opened 2024-06-18
Fix some typos and grammar issues in the README.

rcorre:
This fixes a few spelling and grammar issues I noticed in the README.
Overall I tried not to change the structure too much, but rather made
small tweaks to make it sound a little more correct.

This also removes trailing whitespace from some lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -826,7 +826,7 @@ println!("{}", table);
 An output could look like the following.
 
 ```text
-┌───┬───┬───┐
+┌───┬───┬───┐ 
 │ 0 │ 1 │ 2 │▒
 ├───┼───┼───┤▒
 │ A │ B │ C │▒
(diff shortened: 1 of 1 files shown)


Alexdelia:
need to check if `tabled` prints those whitespaces:

https://github.com/zhiburt/tabled/pull/408/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L366-L369

if it does, then they must not be removed

zhiburt:
Hi @rcorre @Alexdelia

It seems we will fix it forever :sweat_smile: 
Someone always has been bad with grammar :disappointed: 

Thanks a lot.
Sorry for a delay

Take care.
And have a good weekend.

PR #412, opened 2024-06-26
feat: add i18n support

brianheineman:
Add support for customizing the record template for extended tables so that clients (e.g. https://github.com/theseus-rs/rsql) using this library can localize the template.


zhiburt:
Hi @brianheineman 

I tend to think that the original problem is way simpler :smile: just provide a method to replace `"Record"` text.
So you'd be able to use `i18n` on your own by just setting this value.

Something like

```rust
ExtendedTable::new(&data)
    .template(|f, index| write!(f, "[ {} {index} ]", localized("record_name")));
```

What do you think?

And I think we do missing this method, anyway.
Let me know if you'd like to contribute.

PS: Good call with introduction of a feature.

__________________________

Though I am thinking is there somewhere an actual need/feature for `i18n`, we could provide, probably no?

brianheineman:
@zhiburt I really like your suggestion and I updated the PR to reflect the change.  Please let me know if there are any other changes you would like to see.

zhiburt:
Thanks for contribution.

To be honest I doubt when it will be released.
I have a few important changes on the list, which I am struggle to have done for a few month already ..............
Hopefully soon :disappointed: 

Take care.
Have a good weekend.

brianheineman:
Thanks @zhiburt I appreciate the feedback and quick response!

Issue #410, opened 2024-06-20
Wrapping breaks when column contains special chars


madchicken:
When a column text contains special chars, like \n or \t and the column is small enough, the current wrapping algorithm panic since in [wrap.rs at line 309](https://github.com/zhiburt/tabled/blob/master/tabled/src/settings/width/wrap.rs#L309) `line_with` > `width`.
To reproduce the bug, [you use this code](https://gist.github.com/madchicken/1dc891cd5192231729c0f750d84c2a65)
Please, run it in a small terminal (like 80-100 columns) otherwise you won't see the problem.
I think the problem is in the way the width is calculated inside the `split_at_width` function in utils/strings.rs

zhiburt:
Hi @madchicken 

Panics are bad...................................................

> When a column text contains special chars, like \n or \t

`\t`, `\r` in particular will likely break things up
see https://github.com/zhiburt/tabled/tree/master?tab=readme-ov-file#charset

I am sure we shall avoid panic, I'll take a look at your patch.

Thanks for bringing it up;
And sorry for such an issue....

madchicken:
I completely missed that part of the documentation, my bad. My PR is probably to close, but I wanted to give it a try.
In general, it would be good to have a way to split headers predictably. In my case, for example, I have the name of a table and on the next line I am trying to display the name of the column (if present in my data)

madchicken:
I think the example I attached to the issue should let you understand what I mean, where I know the header has a break line: 

https://gist.github.com/madchicken/1dc891cd5192231729c0f750d84c2a65

but I think there is a bigger problem in the fact the code doesn't count "special" chars when trying to split lines in the wrap code (they are ignored since the returned size is always 0).

PR #415, opened 2024-07-11
Fix clippy and tests fail due to simple problems 

### LINKED ISSUE #414: clippy and tests fail due to simple problems
This issue is more of a way to understand the status of this project, as I love it and want to contribute by fixing some bugs. However, I found out that `cargo test` and `clippy` and thus CI don't work due to simple code errors.

Why is everyone ignoring them? Is it intentional or is it okay if someone fixes these errors?

strange-dv:
Closes #414 

This PR fixes small issues that cause CI to fail due to failing tests and `cargo clippy`

## Testing
Local testing using `cargo test --workspace --no-fail-fast --all-features` and `cargo clippy --all --all-targets --all-features`


papergrid/tests/grid/render.rs
```diff
@@ -210,32 +210,33 @@ test_table!(
...
+// #[ignore = "Fails due to the same reason as the previous test"]
+// test_table!(
+//     doesnt_render_return_carige_1,
+//     grid(2, 2).change_cell((1, 1), "12345678").change_cell((0, 1), "123\r\r\r567").build(),
+//     "+---+--------+"
+//     "|0-0|123\r\r\r567  |"
+//     "+---+--------+"
+//     "|1-0|12345678|"
+//     "+---+--------+"
+// );
```
strange-dv, line 239: I had to comment out these lines until the problem described in https://github.com/zhiburt/tabled/issues/410 is solved

tabled/src/lib.rs
```diff
@@ -430,21 +430,6 @@ pub use crate::{tabled::Tabled, tables::Table};
...
-///
-/// #[derive(Tabled)]
-/// struct Record {
-///     #[tabled(skip)]
-///     id: [u8; 4],
-///     #[tabled(format("{}.{}.{}.{}.{}", self.id[0], self.id[1], self.id[2], self.id[3], self.name)]
-///     name: String,
-/// }
-/// ```
-///
```
strange-dv, line 447: I couldn't find anything about `#[tabled(format)]` in the docs and assumed that we do not support it anymore so we do not need tests for it. I would be glad to be proven wrong.

tabled/src/lib.rs
```diff
@@ -439,8 +439,8 @@ pub use crate::{tabled::Tabled, tables::Table};
 /// #[derive(Tabled)]
 /// struct Record {
 ///     #[tabled(skip)]
-///     id: [u8; 4],
-///     #[tabled(format("{}.{}.{}.{}.{}", self.id[0], self.id[1], self.id[2], self.id[3], self.name)]
+///     id: u8,
+///     #[tabled(format("{}.{}", self.id, self.name))]
```
strange-dv, line 443: We do not support formatting using complex expressions like `self.id[0]` so I simplified this code a bit

zhiburt: That's actually one of my endless (not released patches) :smile: 
I did rewrite this support more then a few times.

I wonder what do you think in this regard.
Shall we support functions calls etc?

```rust
#[tabled(format("{}.{}", some_function(self.id), self.name.one_more_field))]
id: u8,
```

Why I did rewrote it is cause it's impossible to support in inline enums :disappointed: 
And having 2 different approaches in 2 different places seeeems like not the right approach.

```rust
#[derive(Tabled)]
enum Something {
#[tabled(inline)]
A {
   #[tabled(format("{}", self.field1, func(0, 1, self)))]  // <- impossible to complex expressions with self 
   field1: usize,
},
B {}
}
```

So anyway maybe you have some thoughts

zhiburt:
Hmmmm

I did not realized it since today.
The errors comes from https://github.com/unicode-rs/unicode-width/issues/65

If on `unicode-width="0.2.11"` it all works but on newest its not.

```log

failures:
    core::table_test::derived::table_emojie_multiline
    settings::render_settings::tab_isnot_handled_by_default_test
failures:
    util::string::tests::strip_test
failures:
    grid::format_configuration::tabs_arent_handled
    grid::render::doesnt_render_return_carige_0
    grid::render::doesnt_render_return_carige_1
    grid::render::hieroglyph_handling_2

```

zhiburt:
I think we need put some thinking to how to resolve it.
Either to force `0.2.11` usage.
Or adjust used logic, accordingly.

tabled/src/lib.rs
strange-dv: What do you mean by rewriting? What do we have now if this is not possible to implement for inline structs? Is this the same functionality I assumed for the test fix? If so, I'm sure that supporting more advanced formatting functionality is a good feature, but obviously, there are many complications that can affect the decision of whether it's worth implementing or not.

If you have patches that extend this functionality beyond the scope of the changed test, what should this test look like to test as much of `#[tabled(format()]` as possible?

strange-dv:
> I think we need put some thinking to how to resolve it.
> Either to force `0.2.11` usage.
> Or adjust used logic, accordingly.

I'm gonna figure out how problematic those adjustments will be and try to fix this issue in this PR since I'm pretty sure enforcing the usage of the outdated version that works for us will cause problems for some users (with `cargo deny` for example).

tabled/src/settings/width/wrap.rs
```diff
@@ -1431,7 +1441,7 @@ mod tests {
                 "\u{1b}[37m(l\u{1b}[39m\u{1b}[37m\u{1b}[41marg\u{1b}[39m\u{1b}[49m\u{1b}[37mest \u{1b}[39m \n",
                 "\u{1b}[37minteger \u{1b}[39m  \n",
                 "\u{1b}[37mless than \u{1b}[39m\n",
-                "\u{1b}[37more equal \u{1b}[39m \n",
+                "\u{1b}[37mor equal \u{1b}[39m \n",
```
Alexdelia, line 1444: that's my fault, sorry guys

Alexdelia: I've re-reviewed my previous `fix typos` `MR` to make sure I didn't make other mistakes and I've reverted those error

At the same time, I fix newly found typos

I hope you pardon my mistake and appreciate the `MR`

#416

tabled_derive/src/parse/field_attr.rs
```diff
@@ -15,6 +15,7 @@ pub fn parse_field_attributes(
 }
 
 pub struct FieldAttr {
+    #[allow(dead_code)]
```
Alexdelia, line 18: why choose to allow `dead code`?

strange-dv: I'd like to hear from @zhiburt on this.
I see this field was added in https://github.com/zhiburt/tabled/commit/ba048dbc22922ae0584a21d83a6ce10f43dd099e. I know it was a long time ago, but do you remember why it exists but never read or it's fine if I remove it?

As for those `#[allow(dead_code)]` that I left in the test code, I think the reason why they are there is more understandable - they are intended for fields that we never read in tests, but should be able to write to them

zhiburt: > As for those #[allow(dead_code)] that I left in the test code, I think the reason why they are there is more understandable - they are intended for fields that we never read in tests, but should be able to write to them

correct;
though these test could be cleaner I guess

> if I remove it?

Will be all right.

zhiburt:
As far as I think about it nowadays could make a 3 changes in `Cargo.toml`s (`papergrid/Cargo.toml`, `tabled/Cargo.toml`, `testing_table/Cargo.toml`)

```diff
-unicode-width = "0.1"
+unicode-width = "=0.1.11"
```

Once you do the failed tests must be fixed on its own.
And we could merge it.

tabled_derive/src/parse/type_attr.rs
```diff
@@ -14,6 +14,7 @@ pub fn parse_type_attributes(
 }
 
 pub struct TypeAttr {
+    #[allow(dead_code)]
```
Alexdelia, line 17: shouldn't we also remove this `dead code`?

strange-dv: Damn, missed it, thanks

Alexdelia: that's what `reviews` are for

Alexdelia (approved):

zhiburt:
OK seems good to me :100: 

Thanks @Alexdelia for review
Thanks for commitment @strange-dv 

Have a good week;
Take care

PR #422, opened 2024-08-11
fix broken link and style tweak

howjmay:


README.md
```diff
@@ -22,93 +22,97 @@ you can find more examples in the **[examples](/tabled/examples/)** folder.
...
-- [Notes](#notes)
-  - [Charset](#charset)
-  - [ANSI escape codes](#ansi-escape-codes)
-  - [Emoji](#emoji)
-  - [Semver](#semver)
-  - [MSRV](#msrv)
-  - [Comparison](#comparison)
+- [    tabled ](#----tabled-)
+  - [Table of Contents](#table-of-contents)
+  - [Usage](#usage)
```
Alexdelia, line 27: unsure about this approach

@howjmay, can you elaborate why this is better than the usual

```md
# title

## subtitle
```

zhiburt: I could guess that it's auto formatting.
I had got it a few times.

@Alexdelia thanks for the look

zhiburt (changes requested):
Hi @howjmay 

Yee the link is indeed seemed to be broken.
I wonder if there's more of them in README.

We could merge it but unsure about this change with TOC.

README.md
zhiburt: I could imagine it's an auto formatting.
Thanks for the look @Alexdelia

PR #424, opened 2024-08-30
Use witdh 1 for special chars when slipping strings

madchicken:
A naive attempt to fix issue #410

The PR adds 1 to the width of a string when a special char is encountered.


madchicken:
Sorry, I had to reopen this PR due to a bad command with the previous one.

zhiburt:
Getting back to this;

I am sure,
We do must not panic/break things in case someone did not prepared his strings in advance.

But I am not sure whether it's the right fix.
Maybe it is.
But somehow I doubt.

If it would be free but this `if` is actually may be costly enough from performance stand point of view.

I guess this issue must be solved on the architecture level --- somehow --- :thinking: :thought_balloon: 

Just left some thoughts here.

Have a great weekend and appreciate your work. 
Let me know what you think.

zhiburt:
Maybe like having a **DIRTY** marker to indicate that we shall do some processing beforehand.

PR #429, opened 2024-10-15
Fix a comment to make doctests happy

joshtriplett:
Without this, `cargo test` says:

> warning: unexpected character `'`

Fix typos at the same time.


zhiburt:
Thanks.

Mailbox refused to work

PR #431, opened 2024-10-18
Fix README.md for json_to_table

akiomik:
`set_object_mode` method has been renamed to `object_orientation` at #321.

https://github.com/zhiburt/tabled/blob/300c07bd53c945671e8115811c1689cb1353feff/json_to_table/src/table/mod.rs#L186-L189


zhiburt:
Thanks once again @akiomik

Issue #435, opened 2024-11-01
Having alternating BorderColor matching alternating Colorization::rows?


k86td:
Hello there. I'm asking because I'm not sure if it's possible with this crate, and I'm still new to Rust. I'm trying to create alternating BorderColor to follow alternating Rows colour. The reason is, I want to create a colourized table with the header underlined. To avoid having a single line underlined going through the header, I'm setting a border of `' '` in the first row, the problem is when I try to colourize, borders are added on all rows which are not colourized using the current row `Colorization`. I've tried setting BorderColor, but I was setting it globally for all rows except the first. The result is half the rows would look okay, but the rows with the alternate colours do not display correctly.

Here's a sample of the code I'm using to display my table:

```rust
#[derive(Debug, Tabled)]
pub struct Expense {
    pub uuid: String,
    pub created: DateTime<Utc>,
    pub modified: DateTime<Utc>,
    pub data: Value,
}

// ...

let expenses = ctx.repo.get_all(25)?;

let underline_style = Color::new(AsciiStyling::Underline, AsciiStyling::Reset);
let display_table = Table::new(expenses)
    .with(Style::empty())
    .with(Colorization::rows([Color::BG_BLACK, Color::BG_WHITE]))
    .with(Colorization::exact([underline_style], Rows::first()))
    .modify(
        Rows::new(0..).not(Rows::first()),
        BorderColor::new().left(Color::BG_BLACK),
    )
    .modify(
        Rows::first().not(Columns::first()),
        Border::new().set_left(' '),
    )
    .to_string();

println!("{}", display_table);
```

Is it possible? I've checked the examples and tried different combination and this is the closest to the actual result I'm looking for. I haven't touched really the iterators yet, maybe it's just about extending `Colorize` to accept `BorderColor`? I will update this issue if I make progress on this. Thanks!

zhiburt:
Hi @k86td 

Sorry for late reply.
I am pretty confident that it's possible to do with `tabled`.
But to be honest I haven't understood the issue properly :(
Could you example once again?

_____________

In mean time, look at my example.
I think it's what you want?
To be honest I quite dissatisfied how complex it is :disappointed: 

It looks like it would great to add some type of iterator callback to modify cells.

```rust
use std::convert::TryFrom;

use owo_colors::OwoColorize;
use tabled::{
    settings::{object::Rows, style::BorderColor, themes::Colorization, Color, Style},
    Table,
};

fn main() {
    let expenses = vec![
[...]

PR #437, opened 2024-11-06
feat/rgb-color

k86td:
ref #435

### DIFF
--- a/tabled/Cargo.toml
+++ b/tabled/Cargo.toml
@@ -303,6 +303,12 @@ name = "theme"
 path = "examples/theme.rs"
 required-features = ["derive", "std"]
 
+[[example]]
+name = "interactive"
+path = "example/interactive.rs"
+required-features = ["derive"]
+
+
 [features]
 default = ["derive", "macros"]
 std = ["papergrid/std"]
--- a/tabled/examples/interactive.rs
+++ b/tabled/examples/interactive.rs
@@ -1,0 +1,135 @@
+use std::{thread::sleep, time::Duration};
+
+use tabled::{
+    settings::{
+        object::{ObjectIterator, Rows},
+        style::BorderColor,
+        themes::Colorization,
+        Color, Style,
+    },
+    Table,
+};
+use tabled_derive::Tabled;
+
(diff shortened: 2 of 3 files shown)


tabled/Cargo.toml
```diff
@@ -303,6 +303,12 @@ name = "theme"
 path = "examples/theme.rs"
 required-features = ["derive", "std"]
 
+[[example]]
+name = "interactive"
+path = "example/interactive.rs"
+required-features = ["derive"]
```
zhiburt, line 309: I was thinking you're new in rust,
you had figured it pretty fast :smile:

tabled/examples/interactive.rs
```diff
@@ -0,0 +1,135 @@
...
+#[derive(Tabled, Clone, Debug)]
+struct Item {
+    name: &'static str,
+    category: &'static str,
+    value: f64,
+}
+
+const CLEAR: &str = "\u{1b}[2J";
+
+type Step = Box<dyn Fn(&mut Table) -> (u64, &mut Table)>;
```
zhiburt, line 23: Ahh kind of hard to understand right?)

Probably will need to refactor it afterall

zhiburt (approved):
Looks good to me @k86td 

Sorry that I did not respond faster.

Take care.
Have a great week.

tabled/Cargo.toml
zhiburt: It's been misspelled "examples/interactive.rs"

zhiburt:
And thanks for example

tabled/examples/interactive.rs
k86td: I agree. I could put the steps in its own struct to try and make it a little bit cleaner if that's what you meant.

Like:
```rust
struct StepResult {
  // could have `Duration` directly here instead of `u64` milliseconds
  duration: u64,
  table: &mut Table,
}

struct Step {
  action: Box<dyn Fn(&mut Table) -> StepResult>
}
```

Not sure if this is syntactically correct, I haven't tried it.

PR #449, opened 2024-11-15
Add license file to testing_table

michel-slm:
Other published crates have license texts, add to this one too so this gets bundled in the next published crate.

```
$ cargo package --list --no-verify | grep LICENSE
LICENSE-MIT
```


zhiburt:
I think you're right.
Thanks.
Have a great weekend.

PR #467, opened 2024-12-13
Update heck crate to 0.5

marxin:


zhiburt:
Thanks for commitment @marxin 
Have a great week, and happy new year (I hope not too soon :smile:)
Take care.

marxin:
Thanks for the review. Speaking about the next release, when can I expect a new one?

zhiburt:
> Speaking about the next release, when can I expect a new one?

Well to be honest I don't know,
generally each one contained some major thing.

I could make a patch bump and release these.
But it would be also not ideal, I guess.

I think need to think :) whether there's some things which could be improved/added.
Currently I have nothing in my mind.

Except to test a `R-tree` vs a list of `HashTable` for settings we have now. I was pondering about it maybe a year ago maybe even longer, but decided that current approach is better.

Essentially we have many calls like these, which I still hope to optimize somehow.....

```rust
if hm.get(Global) {}
or else hm.get(Row) {}
or else hm.get(Column) {}
or else hm.get(Cell) {}
```

zhiburt:
I've just had some examples to run; and it seems promising.
So yes I will try to get a proper impl/testing and see.

And if it's cool we will do it this week or next.
But we will hope to cover it this year :smile: (I just often miss....these deadlines)

marxin:
Good! Anyway, I would appreciate a new release, but I understand you have other stuff to do before making a proper release.

PR #474, opened 2025-01-03
[WIP] Patch/update papegrid spanned config

zhiburt:
Improvement of locality of settings used while rendering.

Logically it must be better but benchmarks showed that when used with `ansi` feature it's slower....
IDK why to be honest must be always faster in my mind...

So yes need to check it further


zhiburt:
Yes just for a history sake I did changes on big tables and it was slower not substantially but still

PR #478, opened 2025-01-14
Fix clippy warnings that are breaking CI

joshtriplett:
- **Use `is_empty`**
- **Use `is_some_and`**


zhiburt:
Thanks.

I wonder what do you think about this ever bumping MSRV.
In general shouldn't we worry about some people using old compilers?:)
Something I was asking myself a few times and haven't came to a good conclusion.

Issue #265, opened 2023-01-15
'Tabled' derive macro with optional fields


danieleades:
For my use-case it is common for fields in a table row to be empty, indicated by an `Optional::None` value. This doesn't seem to be supported, since `Option` doesn't implement `std::fmt::Display`.

Is there anyway around this, or do i need to construct the table manually?

zhiburt:
Hi @danieleades 

If I understood you correctly,
You can use `display_with`.

```rust
use tabled::Tabled;

#[derive(Tabled)]
pub struct MyRecord {
    pub id: i64,
    #[tabled(display_with = "display_option")]
    pub valid: Option<bool>
}

fn display_option(o: &Option<bool>) -> String {
    match o {
        Some(s) => format!("is valid thing = {}", s),
        None => format!("is not valid"),
    }
}
```

There was a though to provide these functions like `display_option`, `display_result` or even consider it a default.
But it was (and still) not clear whereas it's worth it.

Issue #490, opened 2025-01-25
[really good to do] Add a way to map Value into another Value rather then a String


zhiburt:
So we do have `#[tabled(display_with)]` which very good, but it does require us to return String.
Which sometimes is not what we are looking for.

Sometimes it's valuable to convert a value into another value so it could be `#[tabled(inline)]`ed.

Here's a good example from #486

```rust
// we have this structure
struct DbEntry {
    index: usize,
    time: u64,
    text: String,
}

// but we wanna to print the structure
pub struct Entry {
    index: usize,
    #[tabled(inline)]
    time: Date,
    text: String,
}

pub struct Date {
    y: usize,
    m: usize,
    d: usize,
}
```

So the result will be something like

```rust
struct DbEntry {
    index: usize,
    #[tabled(map = "Date::from_u64")]
    #[tabled(inline)]    
    time: u64,
    text: String,
}

pub struct Date {
    y: usize,
    m: usize,
    d: usize,
}
```

I guess we need one more macro helper?

_________________

And actually it's valuable not only for `inline` case,
but we could map and the value could `impl Display` so we would kind be ok with it.

So yessss
Very prominent idea.
I do very like it.

KekmaTime:
@zhiburt can i take up this issue

zhiburt:
Hi @KekmaTime sure,

But it's not the easiest issue to cover,
cause `tabled_derive` is not quite clean.

You would need to take a look at how `display_with` implemented, the difference will be in return type of function and that's it I guess?

If you'll encounter problems please let me know and I'll try to help.

https://github.com/zhiburt/tabled/blob/965a8003f4507d238e41cd471546860b6e5fbb91/tabled_derive/src/lib.rs#L411-L431

KekmaTime:
@zhiburt 👍🏽 for sure

zhiburt:
Hey @KekmaTime 
I was curious a bit and have completed it already.
I hope you'll be not against it.

Take care.
Have a great weekend.

PR #491, opened 2025-01-30
feat: adds display_option_with type attr

georgewhewell:
I have some structs with a bunch of `Option` fields i want to put in the table

i found https://github.com/zhiburt/tabled/issues/265 but it is not ergonomic to put attr on each field, so add type-level attr

### DIFF
--- a/tabled/Cargo.toml
+++ b/tabled/Cargo.toml
@@ -69,6 +69,11 @@ path = "examples/derive/display_with.rs"
 required-features = ["derive"]
 
 [[example]]
+name = "derive_display_with_option"
+path = "examples/derive/display_with_option.rs"
+required-features = ["derive"]
+
+[[example]]
 name = "derive_crate_override"
 path = "examples/derive/crate_override.rs"
 required-features = ["derive"]
--- a/tabled/examples/derive/display_with_option.rs
+++ b/tabled/examples/derive/display_with_option.rs
@@ -1,0 +1,51 @@
+use tabled::Tabled;
+
+#[derive(Debug)]
+pub struct Country {
+    name: String,
+    capital: Option<String>,
+}
+
+fn display_option(opt: &Option<String>) -> String {
+    match opt {
+        Some(val) => val.to_uppercase(),
+        None => "UNKNOWN".to_string(),
+    }
+}
+
+#[derive(Tabled)]
+#[tabled(display_option_with = "display_option")]
+pub struct CountryDisplay {
+    name: String,
+    capital: Option<String>,
+}
+
+impl From<Country> for CountryDisplay {
+    fn from(country: Country) -> Self {
+        Self {
+            name: country.name,
+            capital: country.capital,
+        }
+    }
+}
+
+fn main() {
+    let data = vec![
+        Country {
+            name: "France".to_string(),
+            capital: Some("Paris".to_string()),
+        },
+        Country {
+            name: "Germany".to_string(),
+            capital: Some("Berlin".to_string()),
+        },
+        Country {
+            name: "Unknown".to_string(),
+            capital: None,
+        },
+    ];
+
+    let table_data: Vec<CountryDisplay> = data.into_iter().map(Into::into).collect();
+    let table = tabled::Table::new(table_data);
+    println!("{}", table);
+}
--- a/tabled_derive/src/attributes/type_attr.rs
+++ b/tabled_derive/src/attributes/type_attr.rs
@@ -12,6 +12,7 @@ pub struct TypeAttributes {
     pub inline: bool,
     pub inline_value: Option<String>,
     pub crate_name: Option<String>,
+    pub display_option_with: Option<String>,
 }
 
 impl TypeAttributes {
@@ -52,6 +53,9 @@ impl TypeAttributes {
             }
             TypeAttrKind::RenameAll(lit) => {
                 self.rename_all = Some(CasingStyle::from_lit(&lit)?);
+            }
+            TypeAttrKind::DisplayOptionWith(func) => {
+                self.display_option_with = Some(func.value());
             }
         }
 
--- a/tabled_derive/src/lib.rs
+++ b/tabled_derive/src/lib.rs
@@ -223,7 +223,14 @@ fn info_from_fields(
         headers.push(header);
 
         let field_name_result = field_name(i, field);
-        let value = get_field_fields(&field_name_result, &attributes, fields, field_name);
+        let value = get_field_fields(
+            &field_name_result,
+            &field.ty,
+            &attributes,
+            fields,
+            field_name,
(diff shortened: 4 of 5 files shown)


zhiburt:
Hi @georgewhewell 

Good job. You had put a bit of effort :smile: `tabled_derive` is not the cleanest place currently (like OK but could be better :sweat_smile: ). 

yeeee I do absolutely agree that `Option` type is often used,
And having a macro each time is not convenient, to say the least.
As well as having a boilerplate function you need to create.

I want to say that your idea/implementation is interesting very, but doesn't it feel sort of unnatural?
I mean `Option` type is special for sure, but why not to have something like `#[tabled(display_for(ATypeIsGiven, "a_function_here"))]`.
I am just thinking out loud.
I kind of like it but I do not very like the interface, I mean maybe it hard to read.

Once I was thinking to provide this function for `Option`, but decided not to follow... I had doubts about it.

I wanna here your thoughts here, if you don't mind.

__________________________________

Here is an example what can be done if implement `Tabled` for `Option<T>`.
I think it's a good idea.
And probably must be done?
What do you think?

```rust
use tabled::{Table, Tabled};

#[derive(Tabled)]
struct Repository {
    name: String,
    owner: String,
    #[tabled(inline)]
    head: Option<Commit>,
    #[tabled(inline)]
    updated_at: Option<usize>,
}

#[derive(Tabled)]
struct Commit {
    at: u64,
    hash: String,
}

let repos = vec![
    Repository {
        name: String::from("django-sheets"),
        owner: String::from("georgewhewell"),
        head: None,
        updated_at: None,
    },
    Repository {
        name: String::from("undervolt"),
        owner: String::from("georgewhewell"),
        head: None,
        updated_at: None,
    },
    Repository {
        name: String::from("msp-elixir"),
        owner: String::from("georgewhewell"),
        head: Some(Commit {
            at: 0,
            hash: String::from("4342f01"),
        }),
        updated_at: Some(0),
    },
];

let table = Table::new(repos);

println!("{table}");

// +---------------+---------------+----+---------+-------+
// | name          | owner         | at | hash    | usize |
// +---------------+---------------+----+---------+-------+
// | django-sheets | georgewhewell |    |         |       |
// +---------------+---------------+----+---------+-------+
// | undervolt     | georgewhewell |    |         |       |
// +---------------+---------------+----+---------+-------+
// | msp-elixir    | georgewhewell | 0  | 4342f01 | 0     |
// +---------------+---------------+----+---------+-------+
```

_________

I will notice that #490 sort of related I mean if going with this approach `map` also need to be updated after all.

georgewhewell:
thanks for your detailed response- to be honest i am a "programmer with a deadline"- i didn't think too hard about where to put this, just started hacking after a long day.

your proposal to implement for `Option<T>` seems good, that's what i wanted to achieve-
my use case is that often I am exploring graphql apis, i get huge structs where practically every field is an `Option`, and many mixed types, i just want to dump them into a table for easy viewing

for me the main requirement this PR addresses is that i don't have to care about the field types- i use it with the following `display_option_with`:

```
fn display_option_with(opt: &Option<impl Display>) -> String {
    match opt {
        Some(val) => val.to_string(),
        None => "".to_string(),
    }
}
```

not sure if this was intended in your example, but i think excluding the column altogether if all values are `None` is not so good- can make automated parsing of the table more difficult, tables will be different sized, etc. perhaps can be configurable via `#[tabled(skip_column_all_none)]` or something

zhiburt:
So how about it?

```rust
#[derive(Tabled)]
#[tabled(display_type(Option<String>, "display_option"))]
pub struct CountryDisplay {
    name: String,
    capital: Option<String>,
}
```

I feel like it's better?
Of course we duplicate a type but...we can use any type.

___________________

2 more questions I am interested in your opinion on.

1. Does it make sense to provide a default function for an option?
Something like:

```rust

fn display_option<T>(opt: &Option<T>, value: &str) -> String
where
    T: Display,
{
    match opt {
        Some(val) => val.to_string(),
        None => value.to_string(),
    }
}

#[derive(Tabled)]
#[tabled(display_type(Option<String>, "tabled::util::display::option", "UNKNOWN"))]
pub struct CountryDisplay {
    name: String,
    capital: Option<String>,
}
```

2. What about a name? Maybe it can be just `display` and `display_with` be renamed as well to `display`?

zhiburt:
@georgewhewell Look at this....
Just got this idea today....

I like a lot, and it does solves your issue as well, cause it can work with generics as well now.
See how `#[tabled(display_type(Option, "display_option", "UNKNOWN"))]` `Option` does not have a generic so it will match all options.

```rust
fn display_option<T>(opt: &Option<T>, default: &str) -> String
where
    T: ToString,
{
    match opt {
        Some(val) => val.to_string().to_uppercase(),
        None => default.to_string(),
    }
}

#[derive(Tabled)]
#[tabled(display_type(Option, "display_option", "UNKNOWN"))]
pub struct CountryDisplay {
    name: String,
    capital: Option<String>,
}
```

zhiburt:
In any case I still want to hear your thoughts on a macro name and an utility function.

zhiburt:
And yes one more I have for you...

Maybe you have an idea how we could highlight all these nice things people can do with macros.
Cause I do really think they are very handy.
But the documentation is QUITE big and still not good and still not complete....
So all these nice things often may be lost.....

zhiburt:
And as a side note:
When you map from type A to type B for display you don't necessarily need to call `.collect()`.
You can pass an iterator and everything must work.
In general.

PS: It will be just a bit faster :smile:

zhiburt:
So to sum up:
1. I went with renaming it to `display`
2. added utility functions

But I still wanna hear your opinion

zhiburt:
I've merged things so you can use it on master already.
But I do plan to release it in the coming days.
I just need to fix documentation in a few places.

You can use it as

```rust
use tabled::Tabled;

#[derive(Tabled)]
#[tabled(display(Option, "display_option", "UNKNOWN"))]
pub struct Country {
    name: String,
    capital: Option<String>,
    currency: Option<usize>,
}

fn display_option<T>(opt: &Option<T>, default: &str) -> String
where
    T: ToString,
{
    match opt {
        Some(val) => val.to_string().to_uppercase(),
        None => default.to_string(),
    }
}

fn main() {
    let data = vec![
        Country {
            name: String::from("France"),
            capital: Some(String::from("Paris")),
            currency: Some(0),
        },
        Country {
            name: String::from("Germany"),
            capital: Some(String::from("Berlin")),
            currency: None,
        },
        Country {
            name: String::from("Unknown"),
            capital: None,
            currency: Some(1),
        },
    ];

    let table = tabled::Table::new(data);

    println!("{}", table);
}
```

Once again thanks for a good proposal.
And if you have time please check the questions I've left for your.

Take care.
Have a great week.

PR #493, opened 2025-01-31
Add `#[tabled(display_type)]`

zhiburt:
CONTINUATION OF #491


georgewhewell:
nice, looks good! it works for my use-case, thanks.

PR #501, opened 2025-02-18
Fix wrap emojie issue

zhiburt:


zhiburt:
ref: https://github.com/nushell/nushell/issues/15104

PR #529, opened 2025-04-24
Fix `ansi` feature for `tabled` when compiling to `wasm32-unknown-unknown`

cptpiepmatz:
In this PR, I swapped out the default `runtime-rng` feature of `ahash` in `papergrid` with `compile-time-rng`. This makes it possible to compile to `wasm32-unknown-unknown` without `getrandom` complaining about the missing `js` feature.

I’m not sure `runtime-rng` is even needed here. I wanted to enable it only for certain targets, but thanks to [this long-standing bug](https://github.com/rust-lang/cargo/issues/1197), that’s not possible yet.

Related: https://github.com/nushell/nushell/pull/15138

### DIFF
--- a/papergrid/Cargo.toml
+++ b/papergrid/Cargo.toml
@@ -15,9 +15,17 @@ ansi = ["ansi-str", "ansitok"]
 [dependencies]
 unicode-width = "0.2"
 bytecount = "0.6"
-ahash = { version = "0.8", optional = true }
 ansi-str = { version = "0.9", optional = true }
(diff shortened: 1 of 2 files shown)


zhiburt:
Hey @cptpiepmatz thanks;

You did a good job;
When I had a glimpse at the CI log I immediately decided not to go further :smile: I was not even thinking it's `tabled` related.

I've added wasm32 target on CI as well.

> I’m not sure runtime-rng is even needed here.

Something you make me think about.
What's the point of `randomness` we just need basic int hashes?

PR #537, opened 2025-06-02
tabled: don't pull testing_table until assert feature selected

klensy:
Currently `tabled` pulls `testing_table` dep for `std` and `ansi` features unconditionally. Fixes this, so features for `testing_table` will be used only if `testing_table` enabled.

Spotted when updated to tabled 0.19.


zhiburt:
Hmm interesting
Did not know that

How have you @klensy spotted it? `cargo tree`?

Issue #547, opened 2025-07-20
Display with self no longer compiles


pronebird:
Hi,

Updating from tabled 0.17 to 0.18 I have noticed that display_with renamed to display no longer produces the correct invocation:

```
#[tabled(rename = "Description", display("Self::display_description", self))]
pub description: String,
```

The macro output is:

```rs
Self::display_description(
    &self.description,
    (&self),
)
```

But it should be:

```rs
Self::display_description(&self)
```

zhiburt:
Hi @pronebird 

Yes it was changed to the logic you've described.
As I recall because of the differences between `display = "func"` and `display("func")`.
Previously the arguments were different all though the macro is the same...
So it was decided that the current logic is "more intuitive".

Don't you agree?

PS: IMHO I do agree passing 1 argument with self was more elegant.... but ....

_________________

I am still interested in your opinion but you can bypass it using `format`.

```rust
#[tabled(rename = "Description", format("{}", Self::display_description(self)))]
pub description: String,

```

pronebird:
In my opinion, this was quite elegant before:

```rs
struct SomeType {
  #[tabled(rename = "Description", display("Self::display_description", self))]
  pub description: String,
}

impl SomeType {
    fn display_description(&self) -> String {
        let mut s = "Description: ".to_owned();
        s.push_str(&self.description);
        s
    }
}
```

The new definition looks unusual to me:

```rs
impl SomeType {
    fn display_description(description: &String, _self: &Self) -> String {
        // ...
    }
}
```

Also it's counter-intuitive to me to assume that the property value is passed first instead of following the arguments defined in the macro:

```rs
#[tabled(rename = "Description", display("Self::display_description", self))]
```


But maybe that's because I have seen other crates do it the other way around more often than not. So if `self` was passed first, it would match the instance method and map to `&self` automatically:

```rs
fn display_description(&self, description: &String) -> String
```


However on the other side, if `self` is readily available, I don't see why I'd need `description` as an argument. I can just take it from `self.description`.

This is in my opinion how it should work and what I think looks the best in terms of ergonomics:

1. Standalone functions

```rs
struct MyStruct {
    #[tabled(rename = "ISIN", display = "display_optional_string")]
    name: Option<String>,
}

fn display_optional_string(value: &Option<String>) -> &str {
    value.as_deref().unwrap_or_default()
}
```

2. Static methods:

```rs
struct MyStruct {
[...]

PR #563, opened 2025-11-26
Fix documentation references from display_with to display

### LINKED ISSUE #559: Fix examples/derive/display_with.rs documentation for display_with to display rename
It looks like `display_with` was renamed to `display` in https://github.com/zhiburt/tabled/commit/12e727679685280278bfd3c1caa00f0c9e43b055, but the documentation in the [examples/derive/display_with.rs](https://github.com/zhiburt/tabled/blame/master/tabled/examples/derive/display_with.rs) file still mentions `display_with`:

https://github.com/zhiburt/tabled/blob/d0b8a343ff84f27ec85b04fcdd9af42405953525/tabled/examples/derive/display_with.rs#L1-L11

akiomik:
Fixes #559

The `display_with` attribute was renamed to `display` in commit 12e7276, but the documentation in `examples/derive/display_with.rs` and `README.md` still referenced the old name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1596,7 +1596,7 @@ pub struct Motorcycle {
 
 In the above example, the cc field will be formatted using the specified format string "{} cc", where {} is replaced with the value of cc.
 
-Just like with `display_with` attribute, you can pass arguments for more complex formatting scenarios:
+Just like with `display` attribute, you can pass arguments for more complex formatting scenarios:
 
 ```rust
 use tabled::Tabled;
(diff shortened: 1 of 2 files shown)


zhiburt:
close #559

Thanks @akiomik 

If you were able to understand what `#[tabled(display)]` does.
Maybe you could express your opinion on the next issue.

https://github.com/zhiburt/tabled/issues/547

See, there's 2 main questions there.
1. Shall we support a recursive list of macroses? E.g the next one.
     ```rust
             #[derive(Tabled)]
             struct Person {
                 #[tabled(display = "function1"))]
                 #[tabled(display = "function2"))]
                 #[tabled(display = "function3"))]
                 name: String,
             }
     ```

      Currently we will call the last function. But we could call them one by one and pass the result of the call to the next one.

2. How to better support `self` argument.

akiomik:
Thank you for this great library.

1. When it comes to API design, I prefer the "worse is better" approach. If I were designing this API, I'd stick with the current approach of supporting a single function. From a user's perspective, if they need to compose multiple display functions, they can easily do so within the function they pass to `display`. Supporting a list of functions would make the implementation more complex without offering users much additional value.
2. Ahh... this is a hard one. Let me think about it 😇

PR #572, opened 2026-03-27
feat: add LineText::limit() method

### LINKED ISSUE #524: Add `LineText::limit()`
Currently we can't limit the string size from the end.
We can only manually by truncating it at the point of the call.
Which is not convenient.

The issue is to add such a method.
So we could not go over a set number of characters.

It's not a complex issue.
To figure it out checkout `LineText`.

https://docs.rs/tabled/latest/tabled/settings/style/struct.LineText.html

mvanhorn:
## Summary

Added a `limit` builder method to `LineText` that truncates the rendered text to a maximum number of characters.

## Why this matters

There was no way to cap the length of text placed on a border. If the text exceeded the border width, it overflowed. The `limit` method truncates the text before rendering so it stays within a set character count.

## Changes

- `tabled/src/settings/style/line_text.rs`: Added `limit: Option<usize>` field to `LineText`, a `limit(n)` builder method, and char-based truncation in `create_line()`.

## Testing

All 191 existing tests pass. Added a doctest demonstrating `LineText::new("+-.table", Rows::first()).limit(3)` producing only the first 3 characters on the border. Clippy and rustfmt clean.

Closes #524

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/tabled/src/settings/style/line_text.rs
+++ b/tabled/src/settings/style/line_text.rs
@@ -48,6 +48,7 @@ pub struct LineText<Line> {
     offset: Offset,
     color: Option<ANSIBuf>,
     alignment: Option<Alignment>,
+    limit: Option<usize>,
     line: Line,
 }
 
@@ -85,6 +86,7 @@ impl<Line> LineText<Line> {
             offset: Offset::Start(0),
             color: None,
             alignment: None,
+            limit: None,
         }
     }
 
@@ -153,6 +155,30 @@ impl<Line> LineText<Line> {
     /// ```
     pub fn color(mut self, color: Color) -> Self {
         self.color = Some(color.into());
+        self
+    }
+
+    /// Set a limit on the number of characters displayed.
+    ///
+    /// If the text is longer than the limit, it will be truncated.
+    ///
+    /// ```
+    /// use tabled::{Table, settings::style::LineText, settings::object::Rows};
+    ///
+    /// let mut table = Table::new(["Hello World"]);
+    /// table.with(LineText::new("+-.table", Rows::first()).limit(3));
+    ///
+    /// assert_eq!(
+    ///     table.to_string(),
+    ///     "+-.-----------+\n\
+    ///      | &str        |\n\
+    ///      +-------------+\n\
+    ///      | Hello World |\n\
+    ///      +-------------+"
+    /// );
+    /// ```
+    pub fn limit(mut self, n: usize) -> Self {
+        self.limit = Some(n);
         self
     }
 }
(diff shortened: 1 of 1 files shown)


zhiburt:
Hi @mvanhorn

>  If the text exceeded the border width, it overflowed. 

To be honest.
I was thinking the text would just be shrinked automatically.

I wonder, is it useful?
Maybe.
But if you could enlighten me a little I would be pleased :)

Sorry for late response once again.

mvanhorn:
No worries on timing. The use case is CLI table headers -- when you set a `LineText` on a border, the text can be longer than the column. Right now it overflows rather than being clipped.

Auto-shrink to the border width would be a better default, agreed. `limit()` gives manual control when you want to cap text independently of the column width (e.g., truncating a title to 20 chars regardless of table size).

Would you prefer I add auto-shrink-to-border-width as the default behavior instead? `limit()` could stay as an explicit override for cases where the user wants a fixed cap.

zhiburt:
Sorry for such a delay...
Was busy doing...nothing I guess :(

See @mvanhorn it's already working:

```rust
use std::iter::FromIterator;

use tabled::{
    settings::{object::Rows, split::Split, style::LineText},
    Table,
};

fn main() {
    let mut table = Table::from_iter(['a'..='z']);
    table.with(Split::column(10));

    table.with(LineText::new("Matt Van HornMatt Van HornMatt Van HornMatt Van HornMatt Van HornMatt Van HornMatt Van Horn", Rows::first()));

    println!("{table}");
}
```

```text
Matt Van HornMatt Van HornMatt Van HornMa
| a | b | c | d | e | f | g | h | i | j |
+---+---+---+---+---+---+---+---+---+---+
| k | l | m | n | o | p | q | r | s | t |
+---+---+---+---+---+---+---+---+---+---+
| u | v | w | x | y | z |   |   |   |   |
+---+---+---+---+---+---+---+---+---+---+
```

And if limit is necessary someone could just call `Truncate::truncate` on string and that's it.

What do you think @mvanhorn

mvanhorn:
Oh nice, I didn't realize it already auto-truncates at the border width. That handles the overflow case I was targeting.

The remaining use for `limit()` would be capping text shorter than the table width (e.g., truncating a title to 20 chars regardless of how wide the table is). If that's not a use case you see value in, happy to close this one.

zhiburt:
Thinking about it.
Maybe it's all right.

At least we will have a short hand for manual truncate call.

And it will cover a case where we wanna have exactly column text of certain size, regardless of column width.

So thanks for commitment.
And sorry for delay.

Have a great weekend.

mvanhorn:
Thanks for landing this one.

PR #573, opened 2026-03-27
test: add quickcheck test for wrap character preservation

### LINKED ISSUE #536: Create a quickcheck test for `Width::wrap_text`
Create a test which would check that we provide consistency with wrapping.
Meaning **ALL** original characters are present in the output.

Why it's important,
knowing the algorithm I could imagine that we could skip space `' '` characters sometimes which is not ideal...
And maybe there are more corner cases.

https://github.com/zhiburt/tabled/blob/1ea8b9a53ad298d94dd1d86679fbb95824c7f5bd/tabled/tests/qc/tests/span.rs#L11-L23

mvanhorn:
## Summary

Added a property-based test that verifies the wrapping algorithm preserves all original characters in the output.

## Why this matters

As noted in #536, the wrapping algorithm could potentially drop characters (especially spaces) during line breaking. This quickcheck test generates random ASCII strings at random widths and asserts that every character from the input appears in the wrapped output in order.

## Changes

- `tabled/tests/qc/tests/wrap.rs`: New quickcheck test `qc_wrap_preserves_all_ascii_characters`. Generates printable ASCII strings, wraps at random widths (1+), and checks both `keep_words=true` and `keep_words=false`. Uses ASCII to avoid CJK double-width edge cases where a character wider than the wrap width is replaced.

## Testing

Test passes with `cargo test --test wrap` in the `tabled/tests/qc` directory. Quickcheck runs 100 random cases by default.

Closes #536

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/tabled/tests/qc/tests/wrap.rs
+++ b/tabled/tests/qc/tests/wrap.rs
@@ -1,0 +1,27 @@
+use quickcheck_macros::quickcheck;
+
+use tabled::settings::width::Wrap;
+
+/// Verify that wrapping preserves all original characters (excluding newlines
+/// which are used as line separators by the wrapping algorithm).
+///
+/// Uses ASCII-only strings to avoid CJK double-width edge cases where a
+/// character wider than the wrap width must be replaced.
+#[quickcheck]
+fn qc_wrap_preserves_all_ascii_characters(bytes: Vec<u8>, width: u8) {
+    // Map bytes to printable ASCII + space (0x20..0x7E)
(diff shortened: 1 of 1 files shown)


zhiburt:
Interesting test.

Because there were a number of times were it was not true.
And I was thinking whether it was correct behavior.
Specifically regarding continues ' ' (`whitespaces`).

Thank you @mvanhorn for contribution
Sorry for late response.

mvanhorn:
Thanks for the merge!

PR #575, opened 2026-04-20
docs: remove broken LICENSE symlink from show example (#538)

### LINKED ISSUE #538: broken symlink
https://github.com/zhiburt/tabled/blob/1ea8b9a53ad298d94dd1d86679fbb95824c7f5bd/tabled/examples/show/LICENSE points to a file that does not (or no longer) exists.

mvanhorn:
Closes #538.

\`tabled/examples/show/LICENSE\` points at \`../../LICENSE\`, which resolves to \`tabled/LICENSE\` — a file that no longer exists. Every crate in the workspace now ships its own \`LICENSE-MIT\`, and there's no top-level \`LICENSE\` for the symlink to land on.

Per your reply on the issue ("I think maybe the example shall not have any LICENSE anyhow, right?"), this PR drops the dangling symlink. \`cargo check\` in \`tabled/examples/show\` still succeeds.

🤖 AI-assisted.


zhiburt:
Agree.

Thanks a lot.

mvanhorn:
Thanks for getting the dead symlink out of the example.

PR #577, opened 2026-04-25
switch file that table_width_2 example uses

alexanderkjall:
from ../../CHANGELOG.md to ../README.md, so that it doesn't point outside the crate structure.
 
 Based on: https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/tabled/debian/patches/fix-missing-datafile-in-example.patch
 
 As we are trying to run the testsuite when packaging this in Debian.

### DIFF
--- a/tabled/examples/table_width_2.rs
+++ b/tabled/examples/table_width_2.rs
@@ -4,7 +4,7 @@ use tabled::{
 };
 
 fn main() {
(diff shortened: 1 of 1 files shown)


zhiburt:
Hi @alexanderkjall 

> from ../../CHANGELOG.md to ../README.md, so that it doesn't point outside the crate structure.

Make sense.

PR #585, opened 2026-06-26
Keep combining marks with their base grapheme in Width::wrap

mahirhir:
`Width::wrap` splits a combining character (e.g. the decomposed form of "é") from its base grapheme, because a width-0 char is clamped to width 1, inconsistent with keep_words mode. The fix keeps combining marks with their base (variation selectors / keycaps excluded to avoid emoji regressions). 966 tests green.


mahirhir:
Closing this. I'm not able to verify it properly locally, so I'll take it off your queue rather than leave an untested change open. Thanks.

